PR #8, opened 2020-10-14
[Five 12] Crear Servicio de Artefactos

gperelstein:
Descripción
Se debe crear un servicio de artefactos que realice la comunicación con la base de datos.

Cambios realizados:

- Se crearon los EntityModels de Artifact y ArtifactType

- Se crearon los Models de Artifact y ArtifactType

- Se agregaron los mapeos entre EntityModels.Artifact/Models.Artifact y EntityModels.ArtifactType/Models.ArtifactType en AutoMapper del FRF.Core

- Se creó la interfaz IArtifactsService con los métodos GetAll, Get, Save, Update y Delete

- Se creó el servicio ArtifactsService con los métodos GetAll, Get, Save, Update y Delete

- Se hizo la inyección de dependecias del ArtifactService en el Startup


FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
```diff
@@ -17,6 +17,10 @@ public AutoMapperProfile()
                 .ReverseMap();
             CreateMap<DataAccess.EntityModels.ProjectCategory, Models.ProjectCategory>()
                 .ReverseMap();
+            CreateMap<DataAccess.EntityModels.Artifact, Models.Artifact>()
+                .ReverseMap().ForMember(dest => dest.Id, act => act.Ignore()).ForMember(dest => dest.ArtifactType, act => act.Ignore()).ForMember(dest => dest.Project, act => act.Ignore());
```
matiasbeckerle, line 21: Have you considering using new lines on this scenario? I think it will simplify reading the code. Example:

```csharp
CreateMap<DataAccess.EntityModels.Artifact, Models.Artifact>()
    .ReverseMap()
```

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -0,0 +1,173 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace FRF.Core.Services
+{
+    public class ArtifactsService : IArtifactsService
+    {
+        public IConfiguration Configuration { get; set; }
+        public DataAccessContext DataContext { get; set; }
```
matiasbeckerle, line 16: These two should be `private readonly` instances only (no getter and setters).

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -0,0 +1,173 @@
...
+        private readonly IMapper Mapper;
```
matiasbeckerle, line 17: These 3 above could benefit the .net naming standard as `_name`. Like:

```csharp
private readonly IMapper _mapper;
```

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -0,0 +1,173 @@
...
+                if (result == null)
+                {
+                    return null;
+                }
+                return Mapper.Map<List<Artifact>>(result);
+            }
+            catch (Exception e)
+            {
+                throw e;
+            }
```
matiasbeckerle, line 40: Please read what we have been talking with @gonzalorom in the following thread: https://github.com/MakingSense/five-rocking-fingers/pull/9#discussion_r505692188.

The TL;DR is reduce as much as possible the "exception driven development" and use exceptions for exceptional cases only. This advice applies to all the service methods in this class.

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        public Artifact Save(Artifact artifact)
+        {
+            if (String.IsNullOrWhiteSpace(artifact.Name))
```
matiasbeckerle, line 62: Have you considered using validation attributes instead?

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -0,0 +1,173 @@
...
+                }
+                return Mapper.Map<List<Artifact>>(result);
+            }
+            catch (Exception e)
+            {
+                throw e;
+            }
+        }
+
+        public Artifact Get(int id)
```
matiasbeckerle, line 43: I think you will heavily benefit from using async/await on all these methods. Please consider using the async versions of SingleOrDefault, SaveChanges, etc.

FiveRockingFingers/FRF.DataAccess/DataAccessContext.cs
```diff
@@ -17,6 +19,12 @@ public DataAccessContext(DbContextOptions<DataAccessContext> options) : base(opt
         protected override void OnModelCreating(ModelBuilder builder)
         {
             builder.Entity<ProjectCategory>().HasKey(pc => new { pc.ProjectId, pc.CategoryID });
+            builder.Entity<ArtifactType>().HasKey(at => new { at.Id });
+        }
+
+        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
+        {
+            optionsBuilder.UseSqlServer("server=DESKTOP-VM8EMI2;database=FiveRockingFingers;trusted_connection=true;");
```
matiasbeckerle, line 27: Please consider receiving this value from a configuration.

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
matiasbeckerle: Recommended reading for you both: https://ericlippert.com/2008/09/10/vexing-exceptions/

I've found that post very revealing. Plus, Lippert is always on point.

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gperelstein: Done!

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gperelstein: Done!

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gperelstein: Great recommendation and a great article from Eric Lippert. I will take this recommendations and improve the code! Thank you!

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gperelstein: Thanks, Matias for the comment! I started to read how .net validate attributes and I learn a lot. But the problem is, from what I read, that is quite easy to use this type of validation in the controller (you just have to check if ModelState.IsValid is true) but I don't know how to implement this in the service layer. Do you have a recommendation?

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
matiasbeckerle: Thank you for considering all my comments, glad it helps!

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
matiasbeckerle: Exactly, I wanted you to reach that conclusion. Thanks for absorbing so quickly!

Now that you're aware of validation attributes: have you considered doing validation at the presentation level (controller in your case)? For example you could declare your input DTOs with validation attributes and return a bad request if user isn't providing good input data. That way you work with validated data on the service layer and you only check any weird business rule.

I guess you could also make validations in service layer and data access but so far the validation in presentation has served me enough.

Let me know what do you think. Of course folks I'm not being suuuper picky here so feel free to work one change at a time, specially if you need to make progress. You can work on this later. And I love small PRs.

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -0,0 +1,138 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace FRF.Core.Services
+{
+    public class ArtifactsService : IArtifactsService
+    {
+        private IConfiguration _configuration;
+        private DataAccessContext _dataContext;
```
matiasbeckerle, line 16: These two can (and I think it should) be `readonly` too.

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gperelstein: Your comments were very helpful not just to improve the code, they help me to learn a lot! Yes, I agree, the validation in the controller layer should be enough. I was just wondering if there was any way to implement attributes validation in the service layer because I tend to validate more than is necessary. But I have been working in the controller associate with this service and I'm going to incorporate validation attributes.

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gperelstein: Done!

matiasbeckerle (approved):

PR #9, opened 2020-10-15
[FIVE-4] Crear pagina de detalles de proyecto

gonzalorom:
Descripción
Crear una nueva pagina que muestre los detalles del proyecto seleccionado. 

Cambios realizados:

- Se creo elemento dentro de sidebar para listar los detalles.
- Se re organizo distribucion de carpetas segun arquitectura en Release_0.1.
- Modificacion de ProjectDTO.


FiveRockingFingers/FRF.Web/Controllers/SignInController.cs
```diff
@@ -0,0 +1,42 @@
...
+            var userSignIn = Mapper.Map<UserSignIn>(signInDto);
+            try
+            {
+                var token = await SignInService.SignIn(userSignIn);
+                return Ok(token);
+            }
+            catch (Exception e)
+            {
+                return StatusCode(500, e.Message);
+            }
```
matiasbeckerle, line 39: I noticed that in the endpoints you're doing the same every-time: try to execute an action and then tell users that the server just exploded (500 = internal server error).

I would advice something different. Maybe something you can review later and update the code. My proposal is to simplify the code by removing all these try/catch answer 500.

By removing the try/catch code will be much easier to read.

But why? Well... not only because code simplification. It also means that you're going to delegate the responsibility to the framework (in this case .net). When something goes wrong (remember, these things should only happen from time to time, that's why they're called exceptions) framework will know and automatically return that 500 response for us.

And last but not least, the other reason to remove this code is that you're not doing anything special when an error happens. The idea of catching an exception is to try to do something about. Try to recover from the error. Here we're not recovering.

Hope it helps! Let me know if I can clarify anything.

gonzalorom: First of all, thank you for the feedback this clarify me a lot what I was doing wrong.

Regardless of these endpoints, I was using the exceptions to notify users when something went wrong and handling that on the client-side with a message. For example: "Invalid email or password". This is not exceptional.

I will do what you propose, remove the try/catch and also I'm thinking about this approach:
Return from the service a Tupla of <code><boolean, string></code>, can be <code>(isAuthorize, message)</code>. If is not authorize the controller will return a 401 HTTP status code and a message with the reason otherwise return a 200 HTTP status code and the corresponding id_token and if something exceptional happens leave the responsibility to the framework.

Is this correct way to handle invalid sign in and sign up? 

Thanks again for the feedback!

matiasbeckerle: It depends. I believe each scenario is independent even there are a ton that could be similar. The best advice I can give you here is to rely on HTTP standard by using the status codes as much as you can.

For example, in a controlled scenario (not exceptional) like the one you mentioned (user entered the wrong combination of username & password) you should return something different than 500 because you're in control and you know what happened. A valid response type could be a bad request (400). Unauthenticated (401) is more appropriate when someone is trying to access a protected resource and is unauthenticated. Login or sign in aren't a protected resource, just a public endpoint.

On the other hand to answer your other question you can use tuples or whatever else makes sense for your codebase. Each type has a meaning you can take advantage of.

In the recent past I regret the way I have been building the service responses so in retrospect I would advice creating some kind of multi-purpose response where the operation was successful or not, and then append the success result or an error code/message if not. Of course you can start with the tuple too and see how it evolves, it's totally fair.

Remember that server side messages aren't much useful. Client applications could support more languages or want to show something else to the user (control what they render). I think your server should return error codes instead. Something that frontend can know what it means and act accordingly.

FiveRockingFingers/FRF.Web/Controllers/SignInController.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace FRF.Web.Controllers
+{ 
+    [AllowAnonymous]
+    [Route("api/[controller]")]
+    [ApiController]
+    public class SignInController : ControllerBase
+    {
+        public ISignInService SignInService { get; set; }
+        public IMapper Mapper { get; set; }
```
matiasbeckerle, line 19: These two should not be public and not setters either. I would advice changing them to...

```
private readonly ISignInService _signInService;
private readonly IMapper _mapper;
```

gonzalorom: Done!

FiveRockingFingers/FRF.Web/Controllers/SignInController.cs
gonzalorom: Great explanation. 
I'm going with the tuple solution but with a twist. Instead of returning the error message form server side I will return a 400 HTTP status code and, if all was ok, a 200 with the idToken. So the frontend now will act according to the HTTP status code received.

Thank you again for the feedback.

matiasbeckerle: Np.

FiveRockingFingers/FRF.DataAccess/DataAccessContext.cs
```diff
@@ -18,5 +18,9 @@ protected override void OnModelCreating(ModelBuilder builder)
         {
             builder.Entity<ProjectCategory>().HasKey(pc => new { pc.ProjectId, pc.CategoryID });
         }
+        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
+        {
+            optionsBuilder.UseSqlServer("server=DESKTOP-28EG8E0;database=FiveRockingFingers;trusted_connection=true;");
```
matiasbeckerle, line 23: Consider storing this as a configuration value instead of a hard-coded string.

FiveRockingFingers/FRF.Web.Dtos/AutoMapperProfile.cs
```diff
@@ -1,22 +1,16 @@
...
         {
-            CreateMap<Project, ProjectDto>()
-                .ReverseMap();
-            CreateMap<ProjectCategory, ProjectCategoryDTO>()
-                .ReverseMap();
-            CreateMap<Category, CategoryDTO>()
-                .ReverseMap();
+            CreateMap<Project, ProjectDto>().ReverseMap();
+            CreateMap<ProjectCategory, ProjectCategoryDTO>().ReverseMap();
+            CreateMap<Category, CategoryDTO>().ReverseMap();
```
matiasbeckerle, line 13: Do you map these DTOs to business classes? If not then let's remove the reverse maps.

FiveRockingFingers/FRF.Web.Dtos/Projects/ProjectDTO.cs
```diff
@@ -12,7 +12,8 @@ public class ProjectDto
         public string? Owner { get; set; }
         public string? Client { get; set; }
         public int? Budget { get; set; }
-        #nullable disable
+        public DateTime CreatedDate { get; set; }
+#nullable disable
```
matiasbeckerle, line 16: What's this sorcery? Seems like it can be removed.

FiveRockingFingers/FRF.Web/ClientApp/src/App.test.tsx
```diff
@@ -1,22 +0,0 @@
-import * as React from 'react';
-import * as ReactDOM from 'react-dom';
-import { Provider } from 'react-redux';
-import { MemoryRouter } from 'react-router-dom';
-import App from './App';
-
-it('renders without crashing', () => {
```
matiasbeckerle, line 7: Why test was removed? 😢

FiveRockingFingers/FRF.Web/FRF.Web.csproj
```diff
@@ -11,18 +11,28 @@
...
     <PackageReference Include="Swashbuckle.AspNetCore" Version="5.6.0" />
+    <PackageReference Include="System.IdentityModel.Tokens.Jwt" Version="6.7.1" />
   </ItemGroup>
 
   <ItemGroup>
     <!-- Don't publish the SPA source files, but do show them in the project files list -->
     <Content Remove="$(SpaRoot)**" />
+    <Content Remove="ClientApp\src\commons\NewFolder\**" />
     <None Remove="$(SpaRoot)**" />
+    <None Remove="ClientApp\src\commons\NewFolder\**" />
```
matiasbeckerle, line 35: Are these two lines _NewFolder_ needed?

FiveRockingFingers/FRF.Web/appsettings.Development.json
```diff
@@ -7,6 +7,7 @@
     }
   },
   "ConnectionStrings": {
-    "FiveRockingFingers": "Password=frfAdmin;Persist Security Info=True;User ID=frfAdmin;Initial Catalog=FiveRockingFingers;Data Source=DESKTOP-2LJL8JH\\VAS;MultipleActiveResultSets=True"
+    "FiveRockingFingers":
+      "Password=frfAdmin;Persist Security Info=True;User ID=frfAdmin;Initial Catalog=FiveRockingFingers;Data Source=DESKTOP-28EG8E0;MultipleActiveResultSets=True"
```
matiasbeckerle, line 11: Is this change expected? Please have in mind that passwords **should never** be stored in code repositories.

FiveRockingFingers/FRF.Core/FRF.Core.csproj
```diff
@@ -1,11 +1,14 @@
...
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.1</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="AutoMapper" Version="10.0.0" />
+    <PackageReference Include="Amazon.AspNetCore.Identity.Cognito" Version="1.0.5" />
+    <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="8.0.1" />
+    <PackageReference Include="System.IdentityModel.Tokens.Jwt" Version="6.7.1" />
```
matiasbeckerle, line 11: Why these 3? I don't see any code in this PR that is taking advantage of it 🤔

FiveRockingFingers/FRF.DataAccess/DataAccessContext.cs
gonzalorom: Done!

FiveRockingFingers/FRF.Web.Dtos/AutoMapperProfile.cs
gonzalorom: Those are already part of Release_01. But I can have a look at it.

FiveRockingFingers/FRF.Web.Dtos/Projects/ProjectDTO.cs
gonzalorom: Those are already part of Release_01. But I can have a look at it. Done!

FiveRockingFingers/FRF.Web/ClientApp/src/App.test.tsx
gonzalorom: It's back again :grinning:

FiveRockingFingers/FRF.Web/FRF.Web.csproj
gonzalorom: Removed.

FiveRockingFingers/FRF.Web/appsettings.Development.json
gonzalorom: In this case, is just a locally stored database for dev environment also part of Release_0.1.

matiasbeckerle:
Thank you, please let me know when you push the changes.

FiveRockingFingers/FRF.Core/FRF.Core.csproj
gonzalorom: Just AutoMapper is used for mapping projectDTO and Model/Project. The rest are gone.

gonzalorom:
Changes pushed. Thank you @matiasbeckerle . Always on point with the feedback!

FiveRockingFingers/FRF.Web.Dtos/AutoMapperProfile.cs
matiasbeckerle: Yeah when you have some time review them -- with time I've found that adding the reverse mapping setup could lead to issues because automapper, where things "just works magically". When things "just works" under the hood you could find unplanned inconsistencies.

Instead, what I ended up doing lately is to set the ReverseMap only when I'm coding something where I need it. Otherwise I only map what I need. Hope it helps!

FiveRockingFingers/FRF.Web/FRF.Web.csproj
```diff
@@ -11,10 +11,16 @@
 
   <ItemGroup>
     <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="8.0.1" />
+    <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="8.0.1" />
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="3.1.8" />
```
matiasbeckerle, line 15: `JwtBearer` is used/needed?

FiveRockingFingers/FRF.Web/appsettings.Development.json
matiasbeckerle: I imagined but a good practice is to set your local env hiding the secret values anyway. There's no rush: you can revisit this code later on. Just have it in mind.

matiasbeckerle (approved):

FiveRockingFingers/FRF.Web/FRF.Web.csproj
gonzalorom: Yes in this ticket is not used. Removed.

FiveRockingFingers/FRF.Web/appsettings.Development.json
gonzalorom: I'll search for some examples about it, to learn and improve the code.

FiveRockingFingers/FRF.Web/appsettings.Development.json
```diff
@@ -7,6 +7,7 @@
     }
   },
   "ConnectionStrings": {
-    "FiveRockingFingers": "Password=frfAdmin;Persist Security Info=True;User ID=frfAdmin;Initial Catalog=FiveRockingFingers;Data Source=DESKTOP-2LJL8JH\\VAS;MultipleActiveResultSets=True"
+    "FiveRockingFingers": "Password=frfAdmin;Persist Security Info=True;User ID=frfAdmin;Initial Catalog=FiveRockingFingers;Data Source=DESKTOP-28EG8E0;MultipleActiveResultSets=True",
+    "OnConfiguring": "server=DESKTOP-28EG8E0;database=FiveRockingFingers;trusted_connection=true;"
```
gperelstein, line 11: Maybe the appsettings.Development.json should be in the .gitignore file

gperelstein (approved):
Everything looks great!

martinmena (approved):

FiveRockingFingers/FRF.Web/appsettings.Development.json
gonzalorom: That's a good one. I did not think about this option.  Thank you mate.

PR #11, opened 2020-10-22
[Five 11] Crear API de Artefactos

gperelstein:
Descripción
Se debe crear una API que permita exponga los metodos necesarios para administrar artefactos.

Cambios realizados:

- Se crearon los EntityModels de Artifact y ArtifactType

Se crearon los DTO ArtifactDTO y ArtifactTypeDTO

- Se agregaron los mapeos entre ArtifactDTO/Models.Artifact y ArtifactTypeDTO/Models.ArtifactType en AutoMapper del FRF.Web.Dtos

- Se creó el controller ArtifactsController con los métodos GetAll, Get, Save, Update y Delete

- Se creó una nueva clase padre (BaseApiControllerAsync) para los controladores usando la versión asíncrona de los métodos.


FiveRockingFingers/FRF.Web.Dtos/Artifacts/ArtifactDTO.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace FRF.Web.Dtos.Artifacts
+{
+    public class ArtifactDTO
+    {
+        public int Id { get; set; }
+        [Required(ErrorMessage = "The artifact needs a name")]
+        public string Name { get; set; }
+        public string Provider { get; set; }
+        public XElement Settings { get; set; }
+        public ProjectDto Project { get; set; }
```
matiasbeckerle, line 16: This is somehow minor but consistency is kind of important, specially in projects where you work with other people. I would advice to stick with one approach for naming: using all uppercase DTO or just Dto.

FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
```diff
@@ -0,0 +1,111 @@
...
+        public ArtifactsController(IArtifactsService artifactsService, IMapper mapper)
+        {
+            _artifactsService = artifactsService;
+            _mapper = mapper;
+        }
+
+        [HttpGet]
+        public override async Task<IActionResult> GetAll()
+        {
+            var artifacts = await _artifactsService.GetAll();
```
matiasbeckerle, line 27: Nice, you folks started working with async/await -- this is key! Good job.

FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
```diff
@@ -0,0 +1,111 @@
...
+
+            var artifactsDto = _mapper.Map<IEnumerable<ArtifactDTO>>(artifacts);
+
+            return Ok(artifactsDto);
+        }
+
+        [HttpGet("{id}")]
+        public override async Task<IActionResult> Get(int id)
+        {
+            var artifact = await _artifactsService.Get(id);
```
matiasbeckerle, line 37: Controversial but I will put it on the table. So far .NET team and community appends `Async` at the end (suffix) of every asynchronous method. There's some discussion about whether that was a good or bad idea but so far the suffix is predominant yet until this year.

FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
```diff
@@ -0,0 +1,111 @@
...
+
+            if (artifact == null)
+            {
+                return NotFound();
+            }
```
matiasbeckerle, line 42: 👍 Good use of _what does this value mean?_ to HTTP response.

FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
```diff
@@ -0,0 +1,111 @@
...
+
+            var artifactDto = _mapper.Map<ArtifactDTO>(artifact);
+
+            return Ok(artifactDto);
+        }
+
+        [HttpPost]
+        public override async Task<IActionResult> Save(ArtifactDTO artifactDto)
+        {
+            if (artifactDto == null)
```
matiasbeckerle, line 52: Is this verification needed? I don't remember. I wonder if the attribute validation already makes this fail for you.

FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
```diff
@@ -0,0 +1,111 @@
...
+            {
+                return BadRequest("Artifact object is null");
+            }
+
+            if (!ModelState.IsValid)
```
matiasbeckerle, line 57: Have you tried not using this? I remember someone telling me that in the newer versions of the framework (in my project I'm in 2.2) there was an update where this is triggered automatically so this wasn't necessary anymore. Not sure if that's true or not.

FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
```diff
@@ -0,0 +1,111 @@
...
+            }
+
+            var artifact = _mapper.Map<FRF.Core.Models.Artifact>(artifactDto);
+
+            var artifactCreated = _mapper.Map<ArtifactDTO>(await _artifactsService.Save(artifact));
+
+            return Ok(artifactCreated);
+        }
+
+        [HttpPut]
```
matiasbeckerle, line 69: Hmm I think you need to receive the ID in the URL.

gperelstein: This is really necessary? I have done some projects receiving the ID in the URL, as you say, but when we started this project we agree that the ID will be taken from the parameter passed to the method (in this case the artifact)

FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
gperelstein: You are right with this comment and the previous comment! I already tested it and fixed it in the Save and Update methods

FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
matiasbeckerle: Not strictly necessary but a good practice. Otherwise you're gonna loose the benefits of adhering to the HTTP standard. It's quite a similar scenario to those APIs that defeats the purpose of restful apis. Have you saw those? They always answers 200 OK with a body that has something like...

```json
{
  "result": 404,
  "error": "Not found"
}
```

...like why on earth?! Just return an HTTP 404.

`UPDATE` verbs usually goes alongside a resource name (in your case `artifacts`) and the unique way to identify an instance of that resource (in your case an integer). You will make a semantic api. Same as you do when getting the details of that resource, or when deleting it (you are already doing that!).

You will then not only stick with HTTP standards but also start benefiting from it. For example: are you going to have some sort of authorization rules? Having the ID as part of the URL you could build filters that can allow or deny the access to that resource.

gperelstein: I understand! I never see those APIs that you are talking about but I get the idea. I will fix this in this API and the Categories API

matiasbeckerle: There are memes/jokes about too 😅

EkdzCIQWkAo9LXl

gperelstein: Jajajaj I never see this!

gperelstein: I already fix it!

matiasbeckerle (approved):

FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
```diff
@@ -0,0 +1,96 @@
...
+            return Ok(artifactCreated);
+        }
+
+        [HttpPut("{id}")]
+        public override async Task<IActionResult> Update(int id, ArtifactDTO artifactDto)
+        {
+            if(id != artifactDto.Id)
+            {
+                return BadRequest();
+            }
```
matiasbeckerle, line 65: You should completely ignore the ID from inside the ArtifactDTO. The one to use is the one in the URL, always. In fact, it's a good practice to use a different DTO for input data. A DTO that could skip entirely the ID property.

PR #18, opened 2020-10-30
[FIVE-115 - api / service] Relacionar usuarios registrados con proyectos

gonzalorom:
Detalles
Se deben relacionar los usuarios registrados con los proyectos. Un usario puede tener varios projectos y un projecto puede tener varios usuarios.

Cambios realizados:
- Creación de script SQL create table Users X Project
- Creación Entity UsersByProject
- Se modifica ProjectsService para validar UserId
- Se crea entidad UsersByProject con su respectivo mapper.
- Se modifica ProjectDTO, Models/Project


FiveRockingFingers/FRF.Core/Services/ProjectsService.cs
```diff
@@ -7,165 +7,177 @@
...
+                .Where(up => up.UserId == userId)
+                .Include(pr => pr.Project)
+                .ThenInclude(c => c.ProjectCategories)
+                .ThenInclude(ca => ca.Category)
+                .Include(pp => pp.Project)
+                .SingleOrDefaultAsync(p => p.ProjectId == id);
+            */
+            /*Then delete this*/
+            var project =await _dataContext.Projects.Include(p => p.ProjectCategories).ThenInclude(pc => pc.Category).Include(up=>up.UsersByProject)
+                .SingleOrDefaultAsync(p => p.Id == id);
```
matiasbeckerle, line 106: Minor but remember to be consistent the way you layout the code:

```
var project = await _dataContext.Projects
  .Include(p => p.ProjectCategories)
    .ThenInclude(pc => pc.Category)
  .Include(p=>p.UsersByProject)
  .SingleOrDefaultAsync(p => p.Id == id);
```

FiveRockingFingers/FRF.Web/ClientApp/src/commons/NavMenu.tsx
```diff
@@ -22,12 +23,14 @@ const NavMenu = () => {
...
         getProjects();
         getUser();
-    }, [projects.length]);
+    }, []);
 
+    const { isAuthenticated } = useUserContext();
     const getProjects = async () => {
-        const response = await axios.get("https://localhost:44346/api/Projects/GetAll");
+       //const response = await axios.get("https://localhost:44346/api/Projects/GetAll");
+        console.log(isAuthenticated);
```
matiasbeckerle, line 31: Please remove console logs and commented code.

FiveRockingFingers/FRF.Web/Controllers/ProjectsController.cs
```diff
@@ -5,135 +5,126 @@
...
 using System;
 using System.Collections.Generic;
+using System.Net;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Configuration;
 
 namespace FiveRockingFingers.Controllers
 {
-    public class ProjectsController : BaseApiController<ProjectDto>
+    public class ProjectsController : BaseApiControllerAsync<ProjectDto>
```
matiasbeckerle, line 14: I think there's no need to name the base class with Async suffix, only methods.

FiveRockingFingers/FRF.Core/Services/ProjectsService.cs
```diff
@@ -7,165 +7,180 @@
...
         {
-            Configuration = configuration;
-            DataContext = dataContext;
-            Mapper = mapper;
+            _dataContext = dataContext;
+            _mapper = mapper;
         }
 
-        public List<Project> GetAll()
+        public async Task<List<Project>> GetAllAsync(string userId)
```
martinmena, line 26: Shouldn't this be "GetAllByUserIdAsync"? This method doesn't "GetAll" the projects, but instead gets all the projects associated to that userId.

gonzalorom: That's a fair point.  I think it sounds a bit redundant to use "GetAllByUserIdAsync". Considering that in this application only the projects that are associated with the current user are obtained.

matiasbeckerle: Good. Good debate. I know no one asked my opinion but I will talk anyway 😄

When using typed languages like C# a good way to avoid hyper long names on methods is to use intuition based on input and output arguments. Let's use this one as an example.

Whoevers is gonna call this method, the C# will tell he/her that a userId is required. And that it will return a list of projects. And that is asynchronous. All that.

So the moment you start adding more parameters you then start to see methods like. GetByUserIdAndCategoryId which believe me: it gets worst. You already know that is by user id because it ask you to provide the value.

Methods works like APIs.

Sometimes though, you might need an extra long name or to be more specific but those scenarios comes at a later moment and because you have a ton of code to deal with: which is not this case.

Take this comment also with a grain of salt: you are starting and you need to name things very explicitly so your code is easier to read. We all agree that. It's like a bible in programming. But with the years you start to see patterns and you might be read code easily with shorter method names based on input and output parameters.

TL;DR: always be clear when naming things but also consider in and out parameters.

FiveRockingFingers/FRF.Web/ClientApp/src/commons/NavMenu.tsx
gonzalorom: Done.

FiveRockingFingers/FRF.Web/Controllers/ProjectsController.cs
gonzalorom: Done.

matiasbeckerle:
I think you might have to rebase / fetch since there are some conflicts in the PR.

gonzalorom:
> 
> 
> I think you might have to rebase / fetch since there are some conflicts in the PR.

Done. It was some garbage from the ide.

FiveRockingFingers/FRF.Core/Services/ProjectsService.cs
```diff
@@ -7,165 +7,180 @@
...
-            
+            var result = await _dataContext.UsersByProject
+                .Where(up => up.UserId == userId)
+                .Include(pr => pr.Project)
+                .ThenInclude(c => c.ProjectCategories)
+                .ThenInclude(ca => ca.Category)
+                .Include(pp => pp.Project)
+                .Select(pro => pro.Project).ToListAsync();
+
+            return (result == null) ? null : _mapper.Map<List<Project>>(result);
```
matiasbeckerle, line 36: I think this is not needed. I can be wrong, but when source is null automapper returns null as well.

matiasbeckerle (approved):

FiveRockingFingers/FRF.Core/Services/ProjectsService.cs
gonzalorom: You are right Matias, I will change this for `return _mapper.Map<List<Project>>(result);`

FiveRockingFingers/FRF.Web/Controllers/BaseApiController.cs
```diff
@@ -1,17 +1,19 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
 
 namespace FRF.Web.Controllers
 {
     [ApiController]
     [Route("api/[controller]/[action]")]
-    // [Authorize] TODO:Pending AWS Credentials. Login is bypassed!!! [FIVE-6]
     public abstract class BaseApiController<T> : ControllerBase where T : class
```
gperelstein, line 11: I modified this class to accept two generic classes. One for the Save method and one for the Update method so we can use different DTO's for those two methods. What do you think? You can check in the PR [FIVE-166]-UpsertDTO #22

matiasbeckerle: My personal preference in an scenario like this one is to avoid your custom base class. Let all of your controllers just inherit from what the framework recommends and be consistent about implementing the same style of endpoint (like the get/detail one).

It's more probable that not all your controllers needs exactly the same endpoint every time. Sometimes you might need the delete, sometimes you might not, etc.

Of course, if you have an insert method, follow the same naming, style, etc -- but without inheritance/base class.

gperelstein: Yes, I think that removing the base class is going to be better. We thought of this base class at the beginning but now that the application is growing it's making some troubles

PR #23, opened 2020-11-04
[FIVE-115-front]GR: Relacionar Usuarios Registrados con Projectos

gonzalorom:
- Se modifica front abm projectos, se agregan validaciones.
- Se crea metodo GetUserIdByEmail,
- Modificacion en naming de BaseApiController
- Se borran console.log.


FiveRockingFingers/FRF.Web/ClientApp/src/components/ManageProjectsComponents/EditProject.tsx
```diff
@@ -59,11 +79,70 @@ const EditProject = (props: { project: Project, cancelEdit: any, categories: Cat
...
+            }
+            else {
+                props.openSnackbar("No puede ingresar un campo vacio", "warning");
+                setFieldEmail("");
+            }
+        }
+        else {
+            props.openSnackbar("El dato ingresado no es un email valido", "warning");
+            setFieldEmail("");
+        }
```
matiasbeckerle, line 127: Ok I think you have some work to do here. There are two concerns and I bet the two are related (when you make an improvement on one side it helps on the other):

- Edit and new project are almost identical and there's a fair amount of code duplication.
- This handle user method is doing waaay many things. You will need to divide and conquer. There are several ways but I will not give you the complete answer. I prefer to teach you how to fish rather than give you the fish in the plate. The clue, think about the different aspects you're dealing with here: input validation, server communication, server response handling (and error handling), state, output. Give it a thought and let me know what you think.

FiveRockingFingers/FRF.Web/Controllers/UserController.cs
```diff
@@ -66,5 +66,20 @@ public async Task<IActionResult> Logout()
                 return StatusCode(500, e.Message);
             }
         }*/
+
+
+        // [Authorize] 
+        [HttpGet("getuserid")]
+        public async Task<IActionResult> GetUserId(string email)
```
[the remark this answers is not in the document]
gonzalorom: This endpoint is aimed to get the user id from the given email. It is used to associate users with projects. In the nearby future, we will use a Cognito feature to do that.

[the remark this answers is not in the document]
matiasbeckerle: In order to associate the _current logged in user_ to a project?

FiveRockingFingers/FRF.Web/ClientApp/src/components/ManageProjectsComponents/EditProject.tsx
gonzalorom: Awesome a challenge! Yes, I reckon that this handle method is doing way too many things.
I will begin by separating them into different modules, depending on their responsibilities and relating them through their output arguments. I will also consider the approach of Services for calls to the API.

matiasbeckerle: Excellent. Start with the service. That way you're going to remove some code. Then check this method again after that. Consider removing validation, then check again.

Extra tip: don't do it in an excessive way where you remove almost everything from the method and you end up with 1-liner methods. That's the other extreme where it has it's own disadvantages. Start with service to api and validation. At that point the picture will be clearer.

FiveRockingFingers/FRF.Web/Controllers/UserController.cs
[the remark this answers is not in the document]
gonzalorom: We use another method for that. In this case, we have a "many to many" association between users and projects, so when the user creates a new project he/she can associate different users to that project. So to do that there is a text field in the form to write the email of those users. Asynchronously, the "UserId" of the email is obtained and assigned to the project. All this to improve the user experience and that the user does not receive a message after sending the form with the invalidation of the assigned emails. Could you give me your opinion about it? Could this turn out to be a future security flaw?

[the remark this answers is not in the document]
matiasbeckerle: Ah I understand now. Thanks! The scenario I was imagining it was a security flaw but the one you mention it's not. It could be but that should depend on the business. So I believe you going in the right direction.

Now, what do you think about making this endpoint a little bit future-proof by giving it another name and maybe returning another response?

To begin with: URL. Maybe `/users/search?email=abc`

Then method name: SearchUserAsync?

Then response: UserPublicProfileDto (or something that represents the minimum data to expose like the ID and maybe nickname or similar, could be avatar but not much else). Public-like information only.

That way this endpoint provides the capability of searching users (like when you search one in a social network).

Let me know if it's clear -- I feel like I'm not good with words today, my head is out of place.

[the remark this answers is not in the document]
gonzalorom: I think that would be a great idea. As you said, make the method more future-proof. I will start straight away.
No worries, you have been very clear with the explanation!

FiveRockingFingers/FRF.Web/ClientApp/src/components/ManageProjectsComponents/NewProjectDialog.tsx
```diff
@@ -67,19 +79,49 @@ const NewProjectDialog = (props: { create: boolean, categories: Category[] , fin
...
-                props.openSnackbar("Creaci\u00F3n del proyecto exitosa", "success");
-            } else {
-                props.openSnackbar("Ocurri\u00F3 un error al crear el proyecto", "warning");
+    const handleAddUser = async () => {
+        let userEmail: string | null = "";
+        userEmail = ValidateEmail(email.current?.value as string, emailField, props.openSnackbar);
+
+        if (userEmail != null) {
+            const response = await ProjectService.searchUser(userEmail);
+            switch (response.status) {
```
gperelstein, line 88: I think this is great! You thought about a lot of different possibilities for the status code. I will take some of your ideas

FiveRockingFingers/FRF.Web/ClientApp/src/components/ManageProjectsComponents/NewProjectDialog.tsx
```diff
@@ -67,19 +79,49 @@ const NewProjectDialog = (props: { create: boolean, categories: Category[] , fin
...
+                    setFieldEmail("");
+                    break;
+                case 400:
+                    props.openSnackbar("Ocurri\u00F3 un error al asignar un usuario", "error");
+                    setFieldEmail("");
+                    break;
+                default:
+                    props.openSnackbar("Ocurri\u00F3 un error al asignar un usuario", "error");
+                    setFieldEmail("");
+                    break;
```
gperelstein, line 113: The 400 case and the default case have the same effect. Maybe they can be together in the default case?

gperelstein (approved):
Everything looks great to me!

FiveRockingFingers/FRF.Web/ClientApp/src/components/ManageProjectsComponents/NewProjectDialog.tsx
gonzalorom: You are right mate, 4 lines less of code in a crowded method are always welcomed.

matiasbeckerle (approved):
Good step in the right direction. I still think you can make more improvements but let's move forward one step at a time.

PR #30, opened 2020-11-18
[FIVE-163 - Controller y DTOS] : Refactorizar el código de los controladores y servicios de proyecto

gonzalorom:
Descripción

Se debe refactorizar el código de los controladores y los servicios para:

- Asegurarse que todos utilicen métodos Asíncronos
- Agregar el sufijo Async en el nombre de todos los métodos asíncronos
- Agregar DTO’s de entrada de datos

Realizado:
- Creacion ProjectUpsertDTO.
- Mejoras en ProjectService, se implementan constantes para las llamadas a la api.
- Se mejora la seguridad en ProjectsController.
- Se agregan dataannotation para validar budget y se crea validacion custom para UserId.
- Se mejora metodo save y update de projectService para evitar duplicidad de usuarios.
- Se agrega a appsettings expression regular para validar UserId.
- Se modifica la respuesta en Get y Update cuando no se encuenta el project, de 404 a 204.


FiveRockingFingers/FRF.Core/Models/Project.cs
```diff
@@ -1,16 +1,20 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.ComponentModel.DataAnnotations.Schema;
-using System.Text;
+using System.ComponentModel.DataAnnotations;
+using System.Runtime.Serialization;
 
 namespace FRF.Core.Models
 {
+    [DataContract(IsReference = true)]
```
matiasbeckerle, line 8: What is this?

FiveRockingFingers/FRF.Core/Services/ProjectsService.cs
```diff
@@ -37,16 +37,18 @@ public async Task<List<Project>> GetAllAsync(string userId)
...
             // Gets the categories from the database on a list
             var categoryList = new List<EntityModels.Category>();
             foreach (var category in project.ProjectCategories)
             {
-                categoryList.Add(await _dataContext.Categories.SingleAsync(c => c.Id == category.Category.Id));
+                var categoryToAdd =
+                    await _dataContext.Categories.SingleOrDefaultAsync(c => c.Id == category.Category.Id);
+                if (categoryToAdd == null)
+                {
+                    return null;
```
matiasbeckerle, line 48: Careful. I think you wanted to use `continue` instead so you early return to the next iteration of the loop, right? Otherwise I think you're going to be outside the whole method.

FiveRockingFingers/FRF.Core/Services/ProjectsService.cs
```diff
@@ -57,26 +59,45 @@ public async Task<Project> SaveAsync(Project project)
...
+            foreach (var usersByProject in project.UsersByProject
+                .GroupBy(user => new {user.UserId, user.ProjectId})
+                .Select(g => g.First())
+                .ToList())
+            {
+                var userToAdd = await _dataContext.UsersByProject.FirstOrDefaultAsync(u =>
+                    u.UserId == usersByProject.UserId && u.ProjectId == project.Id);
+                if (userToAdd == null)
+                {
+                    userList.Add(userToAdd);
```
matiasbeckerle, line 75: You're adding a null object here, I think.

FiveRockingFingers/FRF.Core/Services/ProjectsService.cs
```diff
@@ -57,26 +59,45 @@ public async Task<Project> SaveAsync(Project project)
...
+                }
+            }
```
matiasbeckerle, line 77: Similar to what you're doing with categories, the intention of your code is clear and works but quite expensive. You have some alternatives but it's OK if you continue this way. You can have them in mind for next time.

### Alternative 1

Make use of `AnyAsync`. That query alone is faster than fetching the entire object from the database. You don't need any value. You just want to know of there is a user, or if the category exists. If you want, you can compare the generated SQL by EF and notice the differences.

### Alternative 2

You can create a list of all the IDs you need to look for in the database. That way you make a single query to the database which is way more cheaper than X queries (1 per entity you want to check). How? Easy peasy.

```
var usersToLookFor = project.UsersByProject
  .Select(ubp => ubp.UserId)
  .Distintict()
  .ToListAsync();

var usersAlreadyAssociated = await _dataContext.UsersByProject
  .Where(ubp => ubp.ProjectId = project.Id)
  .Where(ubp => usersToLookFor.Contains(ubp.UserId))
  .Select(ubp => ubp.UserId)
  .ToListAsync();

// then at this stage you know who is already in the project
```

Please take into account that the query could be slightly different I'm just making some guessing/assumptions. It's to represent the idea.

### Alternative 3

Maybe this is the most professional one but you need to know exactly how EF works. I believe that if you work with the objects the way EF dictates should be, they're gonna handle all these sneaky things for you. This alone will definitively will cost you some trial/error. I think it's worth it, though.

FiveRockingFingers/FRF.Web.Dtos/Projects/ProjectDTO.cs
```diff
@@ -2,19 +2,19 @@
...
         [Required]
         public int Id { get; set; }
         [Required(ErrorMessage = "Project name is required")]
-        public string? Name { get; set; }
-        public string? Owner { get; set; }
+        public string Name { get; set; }
+        [Required]
+        public string Owner { get; set; }
         public string? Client { get; set; }
+        [Range(1, 1000000000000, ErrorMessage = "Budget have to be greater than 0")]
```
matiasbeckerle, line 17: Since you're using data entry DTOs now -- you still need validation attributes in this one? I don't think so.

FiveRockingFingers/FRF.Web.Dtos/Users/CustomValidator.cs
```diff
@@ -16,7 +16,7 @@ protected override ValidationResult IsValid(object value, ValidationContext vali
                 if (string.IsNullOrWhiteSpace(value as string)) return new ValidationResult("Should not be empty");
 
                 var regexEmailPattern = new Regex(configuration["Regex:EmailPattern"]);
-                
+
```
matiasbeckerle, line 19: Hmm there's a built-in email validation attribute in the framework. Wasn't enough for your use case folks?

FiveRockingFingers/FRF.Web.Dtos/Users/CustomValidator.cs
```diff
@@ -33,11 +33,28 @@ protected override ValidationResult IsValid(object value, ValidationContext vali
...
+                    .GetService(typeof(IConfiguration));
+
+                if (string.IsNullOrWhiteSpace(value as string)) return new ValidationResult("Should not be empty");
+
+                var regexEmailPattern = new Regex(configuration["Regex:UserIdPattern"]);
+
+                if (regexEmailPattern.IsMatch((string) value)) return ValidationResult.Success;
+
+                return new ValidationResult("Is not a correct user id");
+            }
```
matiasbeckerle, line 57: What's the purpose of this validator? Not understanding the idea :/

matiasbeckerle (commented):
I think this PR is waaay too long and modifying all the layers at the same time. I would advice dividing and conquering. Maybe some of my comments help you out with that. Consider one type of change at a time.

FiveRockingFingers/FRF.Web.Dtos/Users/CustomValidator.cs
gonzalorom: Yes, it's true, there is already an email validation attribute in the framework but is not enough, for example, if you insert "a@a" for the built-in validation that is a correct email when is not. For that reason, we use a custom email validation with regex.

gonzalorom:
> 
> 
> I think this PR is waaay too long and modifying all the layers at the same time. I would advice dividing and conquering. Maybe some of my comments help you out with that. Consider one type of change at a time.

I'm steel in the processing of finding to correct balance between a too long PR and perfect size easy going PR, I will start to implement your advice _dividing and conquering_.

FiveRockingFingers/FRF.Web.Dtos/Users/CustomValidator.cs
matiasbeckerle: Hmm weird 🤔

matiasbeckerle:
> I'm still in the processing of finding to correct balance between a too long PR and perfect size easy going PR

Awesome! You have so many options about this one! You could submit only the backend or the frontend (but that's still too large for backend changes, because of the mixture of responsibilities you have modified here). You could submit only 1 endpoint (all layers) at a time. Or 1 service. Or 1 controller. Any option you pick for these changes, will be shorter and easier.

The answer is not straightforward. For example you could submit a 1000 lines modification PR that is super easy to review because it's only about doing a RENAME. But you could also submit a 100 lines PR that's super hard because you modify things that aren't related with each other. For example, a service method about project, a controller endpoint for categories and a DTO class for artifacts. That 100 lines will be WAY harder to review than the rename 1000 lines PR.

It only takes practice. The more you get used to doing code reviews, the more you will notice these things.

You're doing great, keep moving forward 💪

FiveRockingFingers/FRF.Core/Models/Project.cs
gonzalorom: I already remove it, it was a test I was trying to implement.

matiasbeckerle: No problem, but I would recommend closing this PR and submit smaller PRs so we can focus on each part in detail.

FiveRockingFingers/FRF.Core/Services/ProjectsService.cs
gonzalorom: I was trying to get a "bad request" back to the client when he/she submit a category that doesn't exist.

FiveRockingFingers/FRF.Web.Dtos/Projects/ProjectDTO.cs
gonzalorom: You are right. I change it.

FiveRockingFingers/FRF.Web.Dtos/Users/CustomValidator.cs
gonzalorom: Since we receive a string as `UserId` from the client and `UserId` is a GUID we need to validate the format of that string to avoid a request with e.g. _"donbarredora"_. We use a string to avoid any conflict with the frontend.

gonzalorom:
> 
> 
> > I'm still in the processing of finding to correct balance between a too long PR and perfect size easy going PR
> 
> Awesome! You have so many options about this one! You could submit only the backend or the frontend (but that's still too large for backend changes, because of the mixture of responsibilities you have modified here). You could submit only 1 endpoint (all layers) at a time. Or 1 service. Or 1 controller. Any option you pick for these changes, will be shorter and easier.
> 
> The answer is not straightforward. For example you could submit a 1000 lines modification PR that is super easy to review because it's only about doing a RENAME. But you could also submit a 100 lines PR that's super hard because you modify things that aren't related with each other. For example, a service method about project, a controller endpoint for categories and a DTO class for artifacts. That 100 lines will be WAY harder to review than the rename 1000 lines PR.
> 
> It only takes practice. The more you get used to doing code reviews, the more you will notice these things.
> 
> You're doing great, keep moving forward 💪

First workaround. Now in this PR I only leave 1 controller and some minor change on the DTOS. **245 LoC** hope its looks better.

I always look forward to PR as an opportunity to learn. 

Thanks again for your feedback, Matias!:raised_hands:

FiveRockingFingers/FRF.Web.Dtos/Users/CustomValidator.cs
matiasbeckerle: If you use the system type `Guid` anywhere you should be good to go, no need to hack around strings.

matiasbeckerle: Frontend will send those values as _strings_ but the framework/language will resolve them automatically if you use `Guid` type.

PR #35, opened 2020-11-20
[FIVE-163] Validaciones y Tipo de dato GUID

gonzalorom:
- Se cambia tipo de dato UserId por GUID.
- Se valida que el dato no este vacio.
- Se agrega validacion de budge mayor a cero.

### DIFF
--- a/FiveRockingFingers/FRF.Web.Dtos/Projects/UsersByProjectDTO.cs
+++ b/FiveRockingFingers/FRF.Web.Dtos/Projects/UsersByProjectDTO.cs
@@ -8,7 +8,7 @@ namespace FRF.Web.Dtos.Projects
     public class UsersByProjectDTO
     {
         [Required]
-        public string UserId { get; set; }
+        public Guid? UserId { get; set; }
         [Required]
         public int ProjectId { get; set; }
     }
--- a/FiveRockingFingers/FRF.Web/Controllers/ProjectsController.cs
+++ b/FiveRockingFingers/FRF.Web/Controllers/ProjectsController.cs
@@ -1,4 +1,5 @@
-﻿using AutoMapper;
+﻿using System;
+using AutoMapper;
 using FRF.Core.Models;
 using FRF.Core.Services;
 using FRF.Web.Dtos;
@@ -7,6 +8,7 @@ using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Configuration;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
 
 namespace FiveRockingFingers.Controllers
 {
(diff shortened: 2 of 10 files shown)


FiveRockingFingers/FRF.Web.Dtos/Projects/UsersByProjectDTO.cs
```diff
@@ -8,7 +8,7 @@ namespace FRF.Web.Dtos.Projects
     public class UsersByProjectDTO
     {
         [Required]
-        public string UserId { get; set; }
+        public Guid? UserId { get; set; }
```
matiasbeckerle, line 11: Are you sure this one is nullable?

FiveRockingFingers/FRF.Web/Controllers/ProjectsController.cs
```diff
@@ -31,10 +33,10 @@ public class ProjectsController : ControllerBase
 
         //TODO: AWS Credentials, Loggin bypassed.Delete this method and Uncomment GetAll() after do:
         [HttpGet("{userId}")]
-        public async Task<IActionResult> GetAllAsync(string userId)
+        public async Task<IActionResult> GetAllAsync(Guid userId)
         {
-            var currentUserId = _configuration.GetValue<string>("MockUsers:UserId");
-
+            var currentUserId = Guid.Parse(_configuration.GetValue<string>("MockUsers:UserId"));
```
matiasbeckerle, line 38: Please add a TODO line here explaining to use the method argument `Guid userId` instead of this configuration value that is temporal.

FiveRockingFingers/FRF.Web.Dtos/Projects/UsersByProjectDTO.cs
gonzalorom: GUID can not be null, because it is a struct and the default value is "00000000-0000-0000-0000-000000000000".
Dataannotation "**Required**" do something like `isValid: if value != null` so here (what I think and for sure there are more solutions) I can see two workarounds to avoid an empty GUID from the request:
1. Create a Custom Validation for "Required.GUID" for example `isValid: if(UserID != default(GUID))`
2. The itchy one:  first make GUID nullable by using the **?** notation and then, because now UserId is null, use the annotation "**Required**".

I hope that I have been clear in the explanation. What is your opinion? Is there a more professional way to solve this situation?

matiasbeckerle: Ah I think I saw this problem before. It wasn't a guid but was something similar. I think your proposal is fair. We could build a custom validator but maybe is not worthy. Have you tried the following?

```
[Required(AllowEmptyStrings = false)]
public Guid UserId { get; set; }
```

If that doesn't work, your solution is ok.

gonzalorom: I tested it and unfortunately, It doesn't work.

FiveRockingFingers/FRF.Web/Controllers/ProjectsController.cs
gonzalorom: Done.

FiveRockingFingers/FRF.Web.Dtos/Projects/UsersByProjectDTO.cs
matiasbeckerle: Thanks for trying it :)

FiveRockingFingers/FRF.Web/Controllers/ProjectsController.cs
matiasbeckerle: Have you pushed the change? Can't see it yet.

matiasbeckerle (approved):

PR #39, opened 2020-11-26
[NoTicket] Simplify mapper for tests

matiasbeckerle:
## Background

I made a promise to make a quality of life improvement on tests for all the automapper related _inconveniences_. You are currently setting up again and again the mappings (like doing twice the work) so instead with the following changes you will rely on the profiles you already built.

I believe there might be more improvements but this so far will become very handy for you guys. Let's see how it evolves.

## Changes

- Added like a builder. I'm not very fan of static classes but this is one of the scenarios where it's quite useful and risk-free. I will create a mapper for you so have it in mind if you later on add more profiles.
- Updated all test classes so now uses the real deal instead of a mocked instance.
- Updated all tests, removing all the mocked stuff about automapper.

### DIFF
--- a/test/FRF.Web.Tests/Controllers/ArtifactsControllerTests.cs
+++ b/test/FRF.Web.Tests/Controllers/ArtifactsControllerTests.cs
@@ -8,7 +8,6 @@ using Moq;
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 using System.Threading.Tasks;
 using System.Xml.Linq;
 using Xunit;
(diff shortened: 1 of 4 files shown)


gonzalorom (approved):
It works like a charm! Now all the code from the tests looks cleaner.

martinmena (approved):
Seems like a much cleaner solution

matiasbeckerle:
Thanks folks!

gperelstein (approved):
I agree with Gonza and Martin, the tests are easier to read!

PR #41, opened 2020-11-30
[FIVE-190 Back] Mejorar codigo relacion usuarios con projectos

gonzalorom:
### **Tareas Realizadas:**
- Se creó `UserProfile `: contendra los datos relevantes de un usuario para ser expuestos en la UI.
- Se actualizaron returns `GetAll` en `ProjectsControllerTests`.
- Se creó `UserProfileUpsert` para solo solicitar `UserId` al momento de crear un proyecto.
- Ahora Project contiene una lista de `UsersProfile`, en consecuencia, se cambió naming en clases afectadas.


FiveRockingFingers/FRF.Web.Dtos/Projects/ProjectViewModel.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace FRF.Web.Dtos.Projects
+{
+    public class ProjectViewModel
```
matiasbeckerle, line 6: I'm not sure I understand the purpose of this pull request. With your current architecture/design I believe you should only expose DTOs at the API layer. And is the frontend/webapp the one that should create / work with view models. So... what I'm missing? I feel like I'm a little bit lost. Of course you can create new DTOs that are more "view-friendly" but... I suspect something else is going on here.

gonzalorom:
Here what I'm trying to do is, when requesting all projects, return a list that contains: the relevant information of a project as well as the list of users related to it. The problem is that the list of related users `UserByProjectDTO` only contains: `userId` and `projectId`. And what we need is also to obtain the email of these users. To do this (_taking into account that to obtain the emails according to the user's id, you have to make a query to AWS Cognito_) a `UserByProjectViewModel` was created (_perhaps the name can be changed to UserProfile_). 

Maybe this type of object is not technically a ViewModel could be more like a ResourceModel .

Do you think this model should be done on the client side? If so, an endpoint would have to be created that performs the queries with AWS Cognito.

I hope that I have been able to be clear in my explanation.

FiveRockingFingers/FRF.Web.Dtos/Projects/ProjectViewModel.cs
matiasbeckerle: Weird... @gonzalorom you submitted a comment but doesn't exist anymore 🤔

matiasbeckerle: Ping me tomorrow on Slack. I believe you need to build DTOs in a slightly different way -- want that makes it more restful and also serves the your need on the frontend side.

gonzalorom:
_Comment from yesterday_
Here what I'm trying to do is, when requesting all projects, return a list that contains: the relevant information of a project as well as the list of users related to it. The problem is that the list of related users UserByProjectDTO only contains: userId and projectId. And what we need is also to obtain the email of these users. To do this (taking into account that to obtain the emails according to the user's id, you have to make a query to AWS Cognito) a UserByProjectViewModel was created (perhaps the name can be changed to UserProfile).

Maybe this type of object is not technically a ViewModel could be more like a ResourceModel .

Do you think this model should be done on the client side? If so, an endpoint would have to be created that performs the queries with AWS Cognito.

I hope that I have been able to be clear in my explanation.

FiveRockingFingers/FRF.Web/Controllers/ProjectsController.cs
```diff
@@ -35,14 +34,33 @@ public class ProjectsController : ControllerBase
...
+                    var userProfile = new UserProfile
+                    {
+                        Email = await _userService.GetEmailByUserId(uId),
+                        UserId = usersByProjectDto.UserId
+                    };
+                    usersProfile.Add(userProfile);
+                }
+
+                project.UsersProfile = usersProfile;
+            }
```
matiasbeckerle, line 62: I believe you should resolve these loops in the service class. Same for the endpoint below. Controllers should be as thin as possible.

matiasbeckerle (commented):
Good work but not there yet. Left some recommendations.

FiveRockingFingers/FRF.Web.Dtos/Projects/UserProfileDTO.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.ComponentModel.DataAnnotations;
+
+namespace FRF.Web.Dtos.Projects
+{
+    public class UserProfileDTO
+    {
+        [Required]
```
matiasbeckerle, line 8: No need to add this attribute.

matiasbeckerle (approved):

PR #46, opened 2020-12-04
[FIVE-193] Quitar hardcoding de claves

gperelstein:
Descripción:

Se debe eliminar la información sensible de los archivos que van a GitHub.

Cambios realizados:

-Se eliminó el conection string a la base de datos del appsettings.Development
-Se agregó la referencia al secrets.json en FRF.Web.csproj
-Se agregó la indicación de la ruta donde se guarda el secrets.json en el README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,64 @@
 # five-rocking-fingers
 Project estimation system 
+
+## Table of Contents
+
+- [Description](#description)
+- [Structure](#structure)
+- [Project Documentation](#project-documentation)
+- [Tools](#tools)
+- [Team](#team)
+- [Workflow](#workflow)
+
+## Description
+
+Five Rocking Fingers is a software development cost estimation system.
+It was originally conceived as an academic proyect, but it is strongly oriented to a common business need.
+
+It consists on a [ASP.NET Web Development solution](https://visualstudio.microsoft.com/es/downloads/) with a [React](https://reactjs.org/) frontend and a [SQL Server](https://www.microsoft.com/es-es/sql-server/sql-server-downloads) database.
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,2 +1,64 @@
...
+
+Five Rocking Fingers follows [Scrum](https://www.scrum.org/) methodology, organized on 2 weeks [sprints](https://drive.google.com/drive/u/1/folders/1_6KpLydBq47Twx6AwBgNiROTDHJbZCmk).
+
+Planned stories are managed on [Jira](https://makingsense.atlassian.net/secure/RapidBoard.jspa?rapidView=73&projectKey=FIVE&view=planning.nodetail&selectedIssue=FIVE-153&issueLimit=100), where developers can task and estimate the stories.
+
+Then, they can start working on [GitHub repository](https://github.com/MakingSense/five-rocking-fingers) following [Git-Flow](https://datasift.github.io/gitflow/IntroducingGitFlow.html).
+
+## Secret keys
+
+The scret keys are store in:
```
matiasbeckerle, line 61: _stored_

matiasbeckerle (approved):

README.md
gperelstein: Done! Also, I was thinking that we can store the secrets file in the drive of the project so we can have access to the updated file at any moment. If we do that I can add that information to the README. What do you think?

matiasbeckerle: I think that's kind of risky. I would prefer to leave secret values out of any shared drive.

gperelstein: Yes, I think you are right! I won't do it then

PR #49, opened 2020-12-09
[FIVE-125] (subTicket FIVE-222) Crear UnitTests del servicio de categorías

martinmena:
## Tareas realizadas
Creada la clase CategoriesServiceTests y 7 métodos en ella:
  - GetAllAsync_ReturnsList
  - GetAllAsync_ReturnsEmptyList
  - GetAsync_ReturnsCategory
  - GetAsync_ReturnsNull
  - SaveAsync_ReturnsCategory
  - UpdateAsync_ReturnsCategory
  - DeleteAsync_Returns


test/FRF.Core.Tests/Services/CategoriesServiceTests.cs
```diff
@@ -0,0 +1,215 @@
...
+            Assert.Equal(result.ProjectCategories[0].Project.Name, project.Name);
+            Assert.Equal(result.ProjectCategories[0].Project.Budget, project.Budget);
+            Assert.Equal(result.ProjectCategories[0].Project.Id, project.Id);
+        }
+
+        [Fact]
+        public async Task GetAsync_ReturnsNull()
+        {
+            // Arrange
+            var categoryId = 99;
```
matiasbeckerle, line 133: In these particular scenarios I would advice always look for ID `0` so that way you're completely sure it doesn't exist.

test/FRF.Core.Tests/Services/CategoriesServiceTests.cs
```diff
@@ -0,0 +1,215 @@
...
+            var result = await _classUnderTest.SaveAsync(categoryToSave);
+
+            // Assert
+            Assert.IsType<Models.Category>(result);
+
+            Assert.Equal(result.Name, categoryToSave.Name);
+            Assert.Equal(result.Description, categoryToSave.Description);
+            Assert.Empty(result.ProjectCategories);
+        }
+        // The case of saving a category with an empty name is solved in the controller (name can't be null)
```
matiasbeckerle, line 163: I think we can drop this comment.

test/FRF.Core.Tests/Services/CategoriesServiceTests.cs
```diff
@@ -0,0 +1,215 @@
...
+            Assert.Equal(result.Description, category.Description);
+            Assert.Equal(result.Description, updatedCategory.Description);
+
+            Assert.Single(result.ProjectCategories);
+            Assert.Equal(result.ProjectCategories[0].Project.Name, project.Name);
+            Assert.Equal(result.ProjectCategories[0].Project.Budget, project.Budget);
+            Assert.Equal(result.ProjectCategories[0].Project.Id, project.Id);
+        }
+        // The case of updating a non existent category is solved in the controller (an Id that doesn't exist)
+        // The case of updating a category to have an empty name is solved in the controller (name can't be null)
```
matiasbeckerle, line 200: Same for these.

matiasbeckerle (approved):
Solid. Very clean.

test/FRF.Core.Tests/Services/ArtifactsServiceTests.cs
```diff
@@ -23,10 +23,7 @@ public ArtifactsServiceTests()
         {
             _configuration = new Mock<IConfiguration>();
 
-            ContextOptions = new DbContextOptionsBuilder<DataAccessContextForTest>()
-                    .UseInMemoryDatabase(databaseName: "Test")
-                    .Options;
-            _dataAccess = new DataAccessContextForTest(ContextOptions, _configuration.Object);
+            _dataAccess = new DataAccessContextForTest("ArtifactsTestsDB", _configuration.Object);
```
matiasbeckerle, line 26: What if you instead make a new unique one per test? Guid.NewGuid() as db name?

test/FRF.Core.Tests/Services/CategoriesServiceTests.cs
```diff
@@ -0,0 +1,209 @@
...
+        private readonly IMapper _mapper = MapperBuilder.Build();
+        private readonly DataAccessContextForTest _dataAccess;
+        private readonly CategoriesService _classUnderTest;
+        private readonly DbContextOptions<DataAccessContextForTest> contextOptions;
+
+        public CategoriesServiceTests()
+        {
+            _configuration = new Mock<IConfiguration>();
+
+            _dataAccess = new DataAccessContextForTest("CategoriesTestsDB", _configuration.Object);
```
matiasbeckerle, line 25: Same as above... a new one per test? (have in mind the constructor will execute once per each test)

gperelstein: Is it not a good idea to have a different database for every test? I'm working on the project service tests and I'm facing the same problems with all the tests sharing the same database. I was thinking about possible solutions and I thought of the same solution that Martin implemented. Another possible solution is trying to implement what appeared here:
https://xunit.net/docs/shared-context
The idea is to create a class that populated the shared context for every collection of tests and then clean it up after the tests are finish

matiasbeckerle: No, it's not. But it all depends on what you want to achieve. Since these are unit tests the ideal scenario is actually 1 db per test: a complete isolated environment. That's the common scenario for unit tests. Remember: you're mocking everything, you arrange the environment.

Instead, for integration tests the ideal is the opposite: one db for all because the purpose of those it's to be as close as a real environment as possible. It's arguable, though.

Glad you bring the fixtures into the table. It's a useful tool but it should be used wisely. Keep them in mind.

gperelstein: So we can create a new database for these unit tests and use fixtures for integration tests? I already use the approach of creating a new database for every test in the project service tests. I will read more about integration tests

gperelstein (approved):
Everything looks excellent! I test it and it works excellent too!

test/FRF.Core.Tests/Services/CategoriesServiceTests.cs
matiasbeckerle: Or use fixtures to stuff other than database aspects. For integration tests it's likely all your tests will share the same database.

PR #51, opened 2020-12-11
[FIVE-214] (UnitTest) Obtener y Listar Productos de AWS

gonzalorom:
- Se creó UnitTest para AwsArtifactsProviderService.
- Se creó UnitTest para AwsArtifactsProviderController.

### DIFF
--- a/test/FRF.Core.Tests/Services/AwsArtifactsProviderServiceTest.cs
+++ b/test/FRF.Core.Tests/Services/AwsArtifactsProviderServiceTest.cs
@@ -2,8 +2,12 @@
 using FRF.Core.Services;
 using Microsoft.Extensions.Options;
 using Moq;
+using Moq.Protected;
 using System.Collections.Generic;
+using System.Net;
 using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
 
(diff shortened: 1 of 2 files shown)


test/FRF.Core.Tests/Services/AwsArtifactsProviderServiceTest.cs
```diff
@@ -0,0 +1,39 @@
...
+        public async Task GetNamesAsync_ReturnList()
+        {
+            // Arange
+            var handlerMock = new Mock<HttpMessageHandler>(MockBehavior.Strict);
+            
+            // Act
+            var result = await _classUnderTest.GetNamesAsync();
+
+            // Assert
+            var okResult = Assert.IsType<List<KeyValuePair<string, string>>>(result);
```
matiasbeckerle, line 34: Since this is not an http response I would advice using a different naming for the variable.

test/FRF.Core.Tests/Services/AwsArtifactsProviderServiceTest.cs
```diff
@@ -0,0 +1,39 @@
...
+            var awsApiStringBase = new AwsApiStringBase(){UrlApiString = "https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/index.json" };
+            _awsApiString = Options.Create(awsApiStringBase);
+            _classUnderTest = new AwsArtifactsProviderService(_awsApiString);
+        }
+
+        [Fact]
+        public async Task GetNamesAsync_ReturnList()
+        {
+            // Arange
+            var handlerMock = new Mock<HttpMessageHandler>(MockBehavior.Strict);
```
matiasbeckerle, line 28: This is unused.

test/FRF.Core.Tests/Services/AwsArtifactsProviderServiceTest.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    public class AwsArtifactsProviderServiceTest
+    {
+        private readonly IOptions<AwsApiStringBase> _awsApiString;
+        private readonly AwsArtifactsProviderService _classUnderTest;
+
+        public AwsArtifactsProviderServiceTest()
+        {
+            var awsApiStringBase = new AwsApiStringBase(){UrlApiString = "https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/index.json" };
+            _awsApiString = Options.Create(awsApiStringBase);
```
matiasbeckerle, line 20: This is not necessary. The constructor of the service only needs a string with the url.

test/FRF.Core.Tests/Services/AwsArtifactsProviderServiceTest.cs
```diff
@@ -0,0 +1,39 @@
...
+using Xunit;
+
+namespace FRF.Core.Tests.Services
+{
+    public class AwsArtifactsProviderServiceTest
+    {
+        private readonly IOptions<AwsApiStringBase> _awsApiString;
+        private readonly AwsArtifactsProviderService _classUnderTest;
+
+        public AwsArtifactsProviderServiceTest()
```
matiasbeckerle, line 17: The only test in this class is kinda poor so far. I believe you should set a couple of scenarios: like what happens when AWS service responds and what happens when it fails. Let's add some juice.

test/FRF.Core.Tests/Services/AwsArtifactsProviderServiceTest.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            var awsApiStringBase = new AwsApiStringBase(){UrlApiString = "https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/index.json" };
```
matiasbeckerle, line 19: Can you update the host? Since this is just a unit test I think it's safe to use some local naming just in case:

```
https://localhost/offers/v1.0/aws/index.json
```

gonzalorom: Here I'm having a hard time trying to mock HttpClient, I couldn't found a scenario where I can do something like: `_mockHttpClient.Setup(mock => mock.GetStringAsync("url")).ReturnsAsync("test");`.
 
In another hand, I can mock HttpClient (implementing a mock on `HttpResponseMessage`) and get an expected value but only on the test layout, as soon as I try to implement something like what I mentioned above I get: "Unsupported expression...Non-overridable members (here: HttpClient.GetStringAsync)".

So I don't know if is correct to emulate the service method on the test layout or if there is another way to setup HttpClient.GetStringAsync on the service side.

I hope I have been clear with the explanation. Otherwise, we can catch for a call tomorrow.

matiasbeckerle: Show me tomorrow.

test/FRF.Core.Tests/Services/AwsArtifactsProviderServiceTest.cs
gonzalorom: Done.

test/FRF.Core.Tests/Services/AwsArtifactsProviderServiceTest.cs
gonzalorom: Deleted. Done.

test/FRF.Core.Tests/Services/AwsArtifactsProviderServiceTest.cs
gonzalorom: Done.

test/FRF.Core.Tests/Services/AwsArtifactsProviderServiceTest.cs
```diff
@@ -0,0 +1,80 @@
...
+            };
+
+            mockResponse.Content.Headers.ContentType = new MediaTypeHeaderValue("application/json");
+            httpMessage
+                .Protected()
+                .Setup<Task<HttpResponseMessage>>(
+                    "SendAsync",
+                    ItExpr.IsAny<HttpRequestMessage>(),
+                    ItExpr.IsAny<CancellationToken>()
+                )
```
matiasbeckerle, line 54: Have this setup in mind. Maybe there are better ways to mock http responses. I'm not sure now but something to have in mind.

matiasbeckerle (approved):

test/FRF.Core.Tests/Services/AwsArtifactsProviderServiceTest.cs
gonzalorom: I will. Thanks, Matias.

PR #54, opened 2020-12-18
[FIVE-263](DAL) Crear metodo de relaciones de artefactos

gonzalorom:
### Tareas realizadas:
- Se creó método `SetRelation` en `ArtifactsService`, validará que existan los ids de los artefactos a relacionar y que la relación no exista actualmente.
- Se agregan Unit Tests en `ArtifactsServiceTests` para metodo `SetRelationAsync`.

### DIFF
--- a/FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
+++ b/FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
@@ -169,13 +169,9 @@ namespace FRF.Core.Services
             {
                 var innerException = except.InnerException as SqlException;
                 if (innerException != null && (innerException.Number == 2627 || innerException.Number == 2601))
-                {
                     return null;
-                }
-                else
-                {
-                    throw;
-                }
+
+                else throw;
             }
             
             return resultArtifactRelations;
--- a/FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
+++ b/FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
@@ -143,10 +143,10 @@ namespace FRF.Core.Services
             return;
         }
 
-        public async Task<IList<ArtifactsRelation>> SetRelation(IList<ArtifactsRelation> artifactRelations)
+        public async Task<IList<ArtifactsRelation>> SetRelationAsync(IList<ArtifactsRelation> artifactRelations)
         {
             var resultArtifactRelations = new List<ArtifactsRelation>();
-            var dbArtifactsId = _dataContext.Artifacts.Select(a => a.Id).ToList();
+            var dbArtifactsId = await _dataContext.Artifacts.Select(a => a.Id).ToListAsync();
             var artifactsRelationIds = artifactRelations
                 .Select(ar => ar.Artifact1Id)
                 .Concat(artifactRelations.Select(ar=>ar.Artifact2Id));
--- a/FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
+++ b/FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
@@ -8,6 +8,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using Microsoft.Data.SqlClient;
 
 namespace FRF.Core.Services
 {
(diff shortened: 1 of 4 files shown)


FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -141,5 +142,39 @@ public async Task Delete(int id)
             await _dataContext.SaveChangesAsync();
             return;
         }
+
+        public async Task<IList<ArtifactsRelation>> SetRelation(IList<ArtifactsRelation> artifactRelations)
+        {
+            var resultArtifactRelations = new List<ArtifactsRelation>();
+            var dbArtifactsId = _dataContext.Artifacts.Select(a => a.Id).ToList();
```
matiasbeckerle, line 149: This is not async, we should make it async and await it.

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -141,5 +142,39 @@ public async Task Delete(int id)
             await _dataContext.SaveChangesAsync();
             return;
         }
+
+        public async Task<IList<ArtifactsRelation>> SetRelation(IList<ArtifactsRelation> artifactRelations)
```
matiasbeckerle, line 146: Missing async suffix.

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -141,5 +142,39 @@ public async Task Delete(int id)
...
+                await _dataContext.SaveChangesAsync();
+            }
+            catch (DbUpdateException except)
+            {
+                var innerException = except.InnerException as SqlException;
+                if (innerException != null && (innerException.Number == 2627 || innerException.Number == 2601))
+                    return null;
+
+                else throw;
+            }
```
matiasbeckerle, line 175: Hmm what are those errors? They looks quite specific. Any chances to avoid them?

matiasbeckerle (commented):
I would recommend reviewing the service method and ignore for now the controller part. Leave that code out of this PR. Get the service done right first: besides the changes, we should also add the tests for the new service method.

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gonzalorom: Those are duplicated primary key violations errors. I am trying to handle the error without breaking the service so that the user can generate the relationship again. Any suggestions on this?

matiasbeckerle: You have some alternatives. The first two that came to mind:

- Why users are allowed to enter data that will cause this primary key violations? Restrict what the user can do. This requires you to think ways in the UI to prevent from entering data that will be problematic.
- Before doing an insert, check if there is any record that already exists with the combination of values that could cause a problem. Do a famous `.Any` query if you're concerned that this scenario will happen. Of course always have in mind performance. If you need to check several records, do it in any optimal way.

Otherwise I guess you can keep the catch but that type of code usually can be avoided with the two examples above and maybe other alternatives.

gonzalorom: Thanks for the recommendations I'll implement the second one. Regarding the Unit Test of the service and the controller, I was currently working on it. But now first I gonna fix this PR.

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gonzalorom: Done

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gonzalorom: Done.

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
matiasbeckerle: Good! Remember that is always a good idea to submit 1 change and the affected tests at the same time 👌 so that way any merge you do is production-ready. That's a good objective to aim for ;)

PR #57, opened 2020-12-21
[Five 235] (Backend) Listar caracteristicas del artefacto seleccionado en la pagina de creacion de artefactos AWS

gperelstein:
Cambios realizados:

- Se instalaron los paquetes AWSSDK.Core y AWSSDK.Pricing

- Se agregaron dos métodos al servicio AwsArtifactsProvider, uno para traer todos los atributos posibles de un servicio con sus correspondientes valores aceptados. El otro para filtrar ofertas

- Se agregaron los correspondientes métodos en el controller AwsArtifactsProvider

- Se creo una instancia del client del SDK de Amazon en el startup siguiendo el patrón singleton

- Se crearon los modelos PricingTerm y PricingDimension para organizar la información correspondiente a las ofertas


FiveRockingFingers/FRF.Core/Models/PricingDimension.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace FRF.Core.Models
+{
+    public class PricingDimension
+    {
+        public string Unit { get; set; }
+        public string EndRange { get; set; }
+        public string Description { get; set; }
+        public string RateCode { get; set; }
+        public string BeginRange { get; set; }
```
matiasbeckerle, line 13: Are these really strings? Shouldn't some be numbers?

FiveRockingFingers/FRF.Core/Services/AwsArtifactsProviderService.cs
```diff
@@ -14,12 +19,14 @@ public class AwsArtifactsProviderService : IArtifactsProviderService
         private const string OffersCodeIndex = "offers.*.offerCode";
         private readonly AwsPricing _awsPricingOptions;
         private readonly IHttpClientFactory _httpClientFactory;
+        private readonly AmazonPricingClient _client;
```
matiasbeckerle, line 22: _pricingClient

FiveRockingFingers/FRF.Core/Services/AwsArtifactsProviderService.cs
```diff
@@ -52,6 +59,149 @@ public class AwsArtifactsProviderService : IArtifactsProviderService
...
+        public async Task<List<ProviderArtifactSetting>> GetAttributesAsync(string serviceCode)
+        {
+            var attributes = new List<ProviderArtifactSetting>();
+            var response = await _client.DescribeServicesAsync(new DescribeServicesRequest
+            {
+                FormatVersion = "aws_v1",
+                ServiceCode = serviceCode
+            });
+
+            if(response != null)
```
matiasbeckerle, line 71: This code already has a lot of nesting, maybe you can make it better by doing an early return:

```
if (response == null)
  return ...

// your other code
```

FiveRockingFingers/FRF.Core/Services/AwsArtifactsProviderService.cs
```diff
@@ -52,6 +59,149 @@ public class AwsArtifactsProviderService : IArtifactsProviderService
...
+                foreach (var service in response.Services)
+                {
+                    foreach (var attributeName in service.AttributeNames)
+                    {
+                        var attributeKeyValue = new KeyValuePair<string, string>(attributeName, ExtractName(attributeName));
+                        var attribute = new ProviderArtifactSetting();
+                        attribute.Name = attributeKeyValue;
+                        attribute.Values = new List<string>();
+
+                        var response2 = await _client.GetAttributeValuesAsync(new GetAttributeValuesRequest
```
matiasbeckerle, line 82: Maybe this can be encapsulated into a private method so this whole code reads easier? And avoid names like response2.

FiveRockingFingers/FRF.Core/Services/AwsArtifactsProviderService.cs
```diff
@@ -52,6 +59,149 @@ public class AwsArtifactsProviderService : IArtifactsProviderService
...
+                                        PurchaseOption = purchaseOption
+                                    };
+
+                                    pricingDetailsList.Add(pricingTerm);
+                                }
+                            } 
+                        }
+                    }
+                }
+            }
```
matiasbeckerle, line 201: Woah... please consider improving this method with some refactoring mechanisms.

FiveRockingFingers/FRF.Core/Services/AwsArtifactsProviderService.cs
```diff
@@ -52,6 +59,149 @@ public class AwsArtifactsProviderService : IArtifactsProviderService
...
+                                    var pricingDimension = new PricingDimension()
+                                    {
+                                        Unit = (string)termPriceDimension.Value.SelectToken("unit"),
+                                        EndRange = (string)termPriceDimension.Value.SelectToken("endRange"),
+                                        Description = (string)termPriceDimension.Value.SelectToken("description"),
+                                        RateCode = (string)termPriceDimension.Value.SelectToken("rateCode"),
+                                        BeginRange = (string)termPriceDimension.Value.SelectToken("beginRange"),
+                                        Currency = termPriceDimension.Value.SelectToken("pricePerUnit").ToObject<JObject>().Properties().ElementAt(0).Name,
+                                        PricePerUnit = (float)termPriceDimension.Value.SelectToken("pricePerUnit").ToObject<JObject>().Properties().ElementAt(0).Value,
+                                    };
```
matiasbeckerle, line 163: Maybe code blocks like these could be extracted to a private method with a good method name? Something like `var pricingDimension = extractPricingDimension(termPriceDimension);`

FiveRockingFingers/FRF.Web.Dtos/Artifacts/ProviderArtifactSettingDTO.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace FRF.Web.Dtos.Artifacts
+{
+    class ProviderArtifactSettingDTO
```
matiasbeckerle, line 7: public?

matiasbeckerle: Is this used at all?

FiveRockingFingers/FRF.Web/Controllers/AwsArtifactsProviderController.cs
```diff
@@ -30,5 +31,21 @@ public async Task<IActionResult> GetNamesAsync()
...
+            return Ok(attributes);
+        }
+
+        [HttpPost]
+        public async Task<IActionResult> GetProductsAsync(List<KeyValuePair<string, string>> settings, string serviceCode)
+        {
+            var products = await _artifactsProviderService.GetProductsAsync(settings, serviceCode);
+
+            return Ok(products);
+        }
```
matiasbeckerle, line 49: I would recommend skipping endpoints for now on this PR that is big enough already and instead add the missing unit tests of the service provider.

FiveRockingFingers/FRF.Web/Startup.cs
```diff
@@ -70,6 +72,7 @@ public void ConfigureServices(IServiceCollection services)
             
             services.Configure<AwsPricing>(Configuration.GetSection(AwsPricing.AwsPricingOptions));
             services.AddHttpClient();
+			services.AddSingleton(new AmazonPricingClient(Configuration["AwsAccessKeys:AccessKeyId"], Configuration["AwsAccessKeys:SecretAccessKey"], RegionEndpoint.USEast1));
```
matiasbeckerle, line 75: I think you can find a better way to use config values in the startup file.

FiveRockingFingers/FRF.Web/FRF.Web.csproj
```diff
@@ -14,6 +14,7 @@
     <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="8.0.1" />
     <PackageReference Include="Amazon.AspNetCore.Identity.Cognito" Version="1.0.5" />
     <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="8.0.1" />
+    <PackageReference Include="AWSSDK.Core" Version="3.5.1.49" />
```
matiasbeckerle, line 17: Do you need this one here? You're already including in core project, and you use core project on web, right?

FiveRockingFingers/FRF.Core/Models/PricingDimension.cs
gperelstein: BeginRange and EndRange could be numbers but seeing some responses of this API, I realize that sometimes EndRange could be something like "inf". The problem is that the documentation of the amazon pricing API is a bit incomplete and not specify the response that you are gonna get in detail

gperelstein: In all my tests BeginRange is a number but I don't know if this variable could be something else.

FiveRockingFingers/FRF.Core/Services/AwsArtifactsProviderService.cs
gperelstein: Done!

FiveRockingFingers/FRF.Core/Services/AwsArtifactsProviderService.cs
gperelstein: Done!

FiveRockingFingers/FRF.Core/Services/AwsArtifactsProviderService.cs
gperelstein: Done!

FiveRockingFingers/FRF.Core/Services/AwsArtifactsProviderService.cs
gperelstein: Yes, I agree that this function was too long and quite difficult to read. I have tried to improve it, I hope that this is going in the right direction and is better now.

FiveRockingFingers/FRF.Core/Services/AwsArtifactsProviderService.cs
gperelstein: Done!

FiveRockingFingers/FRF.Web/Controllers/AwsArtifactsProviderController.cs
gperelstein: I remove the endpoints and the DTO from this PR. I will create a new PR for this controller. I also incorporated the tests of the services that have been modified in this PR to address a recommendation that you gave to Gonzalo Romero

FiveRockingFingers/FRF.Web/Startup.cs
gperelstein: I check the documentation (https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-3.1). There is a method call GetValue, that is the better way to access config values?

gonzalorom: We are already using a POCO to do that in Release_01 it's called `awsCognitoCredential`. With that you can access to `awsCognitoCredential.AccessKeyId` and so on.

FiveRockingFingers/FRF.Core/Models/PricingDimension.cs
matiasbeckerle: Is this class going to represent your business or AWS? Your business or internal classes should be agnostic to whatever the providers use. If your business representation doesn't match with providers, you can map, transform, convert. Does that answer? Always try to use whatever is the best value type for your business models.

If this class represents an AWS response then maybe you should have a deeper look and find out what is this type. Also have in mind that the whole source code of AWS SDK is on github, available. I always find things searching on their repositories.

FiveRockingFingers/FRF.Core/Services/AwsArtifactsProviderService.cs
matiasbeckerle: Yes, reads easier now, thank you!

FiveRockingFingers/FRF.Web/Startup.cs
matiasbeckerle: Not exactly @gperelstein -- you can use that method but there are simpler ways. The one I have in mind is as @gonzalorom mentions, using configuration classes. So for example you could have AwsPricingOptions as a POCO class that holds the config values you need. Then in the startup you *bind* the values and you should be able to use it here.

Example...

```
var awsPricingOptions = Configuration.GetSection("AwsPricingOptions").Get<AwsPricingOptions>();
```

Something like that, could vary.

matiasbeckerle: Precise example (goes to an anchor): https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-5.0#bind-hierarchical-configuration-data-using-the-options-pattern

FiveRockingFingers/FRF.Core/Models/PricingDimension.cs
gperelstein: This should represent our business. I will find the proper way to manage the data that AWS provides. Checking AWS SDK source code is a great idea! Thank you for the recommendation!

FiveRockingFingers/FRF.Web/Startup.cs
gperelstein: Done! I think this way, the code, is cleaner and easier to maintain. Thank you for the recommendation!

FiveRockingFingers/FRF.Core/Base/AwsSdkOptions.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace FRF.Core.Base
+{
+    public class AwsSdkOptions
+    {
+        public const string AwsSDK = "AwsSDK";
```
matiasbeckerle, line 9: In this case I think I prefer _AwsSdk_ for both the name and the value but... personal taste.

FiveRockingFingers/FRF.Web/FRF.Web.csproj
matiasbeckerle: Don't forget to review this ☝️

FiveRockingFingers/FRF.Web/Startup.cs
```diff
@@ -62,8 +63,13 @@ public void ConfigureServices(IServiceCollection services)
 			services.AddSingleton(cognitoUserPool);
 			services.AddCognitoIdentity();
 			//End Cognito
-            
-            services.Configure<AwsPricing>(Configuration.GetSection(AwsPricing.AwsPricingOptions));
+
+			//Start AWS SDK Access Keys
+			var awsSdk = Configuration.GetSection(AwsSdkOptions.AwsSDK).Get<AwsSdkOptions>();
+			services.AddSingleton(new AmazonPricingClient(awsSdk.AccessKeyId, awsSdk.SecretAccessKey, RegionEndpoint.USEast1));
```
matiasbeckerle, line 69: Just checking: singleton, you sure? I think that for anything about AWS I would prefer scoped... or even transient but not sure. Maybe there's a recommendation on the community?

matiasbeckerle (approved):

FiveRockingFingers/FRF.Web/FRF.Web.csproj
gperelstein: I don't understand exactly why but if I don't have this I can't compile the solution and I get an error that said that there is a compatibility problem with AWSSDK.Core and to solve it I have to install AWSSDK.Core in FRF.Web

matiasbeckerle: Ok, let's move forward with it.

FiveRockingFingers/FRF.Web/Startup.cs
gperelstein: I checked for some recommendations about this and I found that Amazon provides its own method for dependency injection called AddAwsService (I already change the code to use it). But seeing the code, I realize that the lifetime of the dependency is Singleton as default:

`public static IServiceCollection AddAWSService<T>(this IServiceCollection collection, AWSOptions options, ServiceLifetime lifetime = ServiceLifetime.Singleton) where T : IAmazonService;`

Why is better for you to use scoped or transient in this case? I can change the lifetime in the method or directly inject the dependency with AddTransient or AddScoped methods but I don't understand the advantages of using these two lifetimes in this scenario

FiveRockingFingers/FRF.Core/Base/AwsSdkOptions.cs
gperelstein: Done! You are right, I think what you propose is better

FiveRockingFingers/FRF.Web/Startup.cs
matiasbeckerle: Hmm interesting. I don't have a strong argument just that the sensation that as singleton the service will be up the first time the app is deployed until it gets down. Scoped, instead, a new instance is created per each request. And with transient (maybe this is too much) a new one every time no matter what. Probably I will still recommend starting with Scoped and see how it behaves.

gperelstein: Perfect, I already change it to scoped. Thank you for the recommendation!

PR #60, opened 2020-12-28
[FIVE-215] Mejorar el código del servicio de artefactos

martinmena:
## Tareas realizadas

- Modificada la interfaz IArtifactsService para que el método delete devuelva una Task<Artifact>.
- Simplificados chequeos de existencia de projectos y tipos de artefactos al crear/modificar artefactos, y la existencia del proyecto al obtener el listado de artefactos asociados. En lugar de obtener el objeto de la base de datos (y en muchos casos con las relaciones a otras tablas incluidas) se verifica que exista algún elemento con el Id al cual se está intentando asociar.
- Simplificada la asociación entre artefactos con los proyectos y los tipos de artefactos en los métodos Save y Update.
- Modificado método Delete para devolver nulo en caso de no existir el artefacto a eliminar, o el artefacto eliminado en caso exitoso.
- Modificados tests del método Delete acorde a las modificaciones en el servicio.

### DIFF
--- a/FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
+++ b/FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
@@ -27,26 +27,17 @@ namespace FRF.Core.Services
         public async Task<List<Artifact>> GetAll()
         {
             var result = await _dataContext.Artifacts.Include(a => a.ArtifactType).Include(a => a.Project).ThenInclude(p => p.ProjectCategories).ThenInclude(pc => pc.Category).ToListAsync();
-            if (result == null)
-            {
-                return null;
-            }
             return _mapper.Map<List<Artifact>>(result);
         }
 
         public async Task<List<Artifact>> GetAllByProjectId(int projectId)
         {
-            var project = await _dataContext.Projects.Include(p => p.ProjectCategories).ThenInclude(pc => pc.Category).SingleOrDefaultAsync(p => p.Id == projectId);
-            if (project == null)
+            if (!await _dataContext.Projects.AnyAsync(p => p.Id == projectId))
             {
                 throw new System.ArgumentException("There is no project with Id = " + projectId, "projectId");
             }
 
             var result = await _dataContext.Artifacts.Include(a => a.ArtifactType).Include(a => a.Project).ThenInclude(p => p.ProjectCategories).ThenInclude(pc => pc.Category).Where(a => a.ProjectId == projectId).ToListAsync();
-            if (result == null)
-            {
-                return null;
-            }
             return _mapper.Map<List<Artifact>>(result);
         }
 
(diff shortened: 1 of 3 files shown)


gperelstein (approved):
Everything looks great to me!

gonzalorom (approved):
Nice performance improvement.

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -27,26 +27,17 @@ public ArtifactsService(IConfiguration configuration, DataAccessContext dataCont
...
             return _mapper.Map<List<Artifact>>(result);
         }
 
         public async Task<List<Artifact>> GetAllByProjectId(int projectId)
         {
-            var project = await _dataContext.Projects.Include(p => p.ProjectCategories).ThenInclude(pc => pc.Category).SingleOrDefaultAsync(p => p.Id == projectId);
-            if (project == null)
+            if (!await _dataContext.Projects.AnyAsync(p => p.Id == projectId))
             {
                 throw new System.ArgumentException("There is no project with Id = " + projectId, "projectId");
```
matiasbeckerle, line 37: This sounds like too much. Throwing an exception because there's not a project? Please have in mind that [exceptions only should be used for exceptions](https://ericlippert.com/2008/09/10/vexing-exceptions/).

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -62,16 +53,12 @@ public async Task<Artifact> Get(int id)
...
                 throw new System.ArgumentException("There is no project with Id = " + artifact.ProjectId, "artifact.ProjectId");
             }
 
-            //Gets the artifactType associated to it from the database
-            var artifactType = await _dataContext.ArtifactType.SingleOrDefaultAsync(at => at.Id == artifact.ArtifactTypeId);
-            if (artifactType == null)
+            if(! await _dataContext.ArtifactType.AnyAsync(at => at.Id == artifact.ArtifactTypeId))
             {
                 throw new System.ArgumentException("There is no ArtifactType with Id = " + artifact.ArtifactTypeId, "artifact.ArtifactTypeId");
             }
```
matiasbeckerle, line 64: Same as above. In a follow up PR maybe you can improve this code by NOT using exceptions on scenarios like these.

matiasbeckerle (approved):

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
martinmena: Would you suggest to return just an empty array, regardless if it's a project with no artifacts, or a projectId that doesn't exist?

matiasbeckerle: I think it depends. For this particular case a "save artifact" method you can go simple with "null" which you can handle on the controller side and assume if you get null it's because something went wrong and return bad request.

If you need more precision about what happened then you could start considering to build some kind of service response like `Result<Artifact>` where `Result` is a new class you create with some useful information like...

```
bool Success
Error Error
T Value
```

And Error is another new class with maybe something like...

```
int Code
string Message
```

Value will be filled only when success is true. Error only when success is false. I don't know, just a simple idea.

matiasbeckerle: Error codes as numbers are always useful because your consumers can use them so they can act accordingly. Error message as string is just a helpful message for whoever is building the consumer side, not a user-friendly message. Besides, each consumer app probably needs localization or have some sort of specific language to communicate to their users. Backend error messages should be never be presented to end users.

egutbrod: Good call Mati, we created a story FIVE-266 to address that. Thanks!

matiasbeckerle: Awesome, thanks folks!

PR #64, opened 2021-01-06
[FIVE-265] Mejorar manejo de Status codes del lado del front

gonzalorom:
### Descripción
Cuando se hacen llamadas a la API con axios y la respuesta es distinta a un HTTP Status code 200, se produce una excepción del lado del servicio (front), haciendo que la evaluación de HTTP Status code no se pueda realizar.
### Acciones realizadas:
- Se mejora respuesta de Promises en llamadas a la API del lado de los servicios del frontend.

### DIFF
--- a/FiveRockingFingers/FRF.Web/ClientApp/src/services/ArtifactService.ts
+++ b/FiveRockingFingers/FRF.Web/ClientApp/src/services/ArtifactService.ts
@@ -9,7 +9,7 @@ class ArtifactService {
         try {
             return await axios.get(`${ARTIFACTS_URL}GetAll`);
         } catch (error) {
-            return error.response ? error.response : null;
+            return error.response ? error.response : error.message;
         }
     };
 
@@ -17,7 +17,7 @@ class ArtifactService {
         try {
             return await axios.get(`${ARTIFACTS_URL}GetAllByProjectId/${id}`);
         } catch (error) {
-            return error.response ? error.response : null;
+            return error.response ? error.response : error.message;
         }
     };
 
@@ -25,7 +25,7 @@ class ArtifactService {
         try {
             return await axios.get(`${ARTIFACTS_URL}Get/${id}`);
         } catch (error) {
-            return error.response ? error.response : null;
+            return error.response ? error.response : error.message;
         }
     };
 
@@ -33,7 +33,7 @@ class ArtifactService {
         try {
             return await axios.post(`${ARTIFACTS_URL}Save`, artifact);
         } catch (error) {
-            return error.response ? error.response : null;
+            return error.response ? error.response : error.message;
         }
     };
 
@@ -41,7 +41,7 @@ class ArtifactService {
         try {
             return await axios.put(`${ARTIFACTS_URL}Update/${id}`, artifact);
         } catch (error) {
-            return error.response ? error.response : null;
+            return error.response ? error.response : error.message;
         }
     };
 
@@ -49,7 +49,7 @@ class ArtifactService {
         try {
             return await axios.delete(`${ARTIFACTS_URL}Delete/${id}`);
         } catch (error) {
-            return error.response ? error.response : null;
+            return error.response ? error.response : error.message;
         }
     };
 }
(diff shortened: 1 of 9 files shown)


FiveRockingFingers/FRF.Web/ClientApp/src/services/ArtifactService.tsx
```diff
@@ -6,33 +6,86 @@ const ARTIFACTS_URL = `${BASE_URL}api/Artifacts/`
...
+        return await axios.get(`${ARTIFACTS_URL}GetAll`).then(response => {
+            if (response.status === 200) {
+                return response;
+            }
+        }
+        ).catch(function (error) {
+            if (error.response) {
+                return error.response;
+            }
+        });
```
matiasbeckerle, line 18: I think that when you use `await` is better to avoid using .then and .catch. Otherwise you may end up working with several nested promises making the code harder to read and debug.

Contrary to the counterpart in C# I think that in this case the community uses try {} catch {} when using `await`. Check online but I guess it's something like that.

FiveRockingFingers/FRF.Web/ClientApp/src/services/ArtifactService.tsx
```diff
@@ -6,33 +6,86 @@ const ARTIFACTS_URL = `${BASE_URL}api/Artifacts/`
...
+        .then(response => {
+            if (response.status === 200) {
+                return response;
+            }
+        }
+        ).catch(function (error) {
+            if (error.response) {
+                return error.response;
+            }
+        });
```
matiasbeckerle, line 32: Maybe in the future -- I can't help you much with JS/TS code right now but you might be able to create some kind of method that does all this in one place and you reuse it everywhere so you avoid duplicating these lines over and over again. I think it could be challenging at this stage so... better gain some knowledge while you build this project then later in the future you will see the way to improve and reuse this code.

FiveRockingFingers/FRF.Web/ClientApp/src/services/ArtifactService.tsx
gonzalorom: Here we have two approaches:
One is use **_.catch_** to catch the error asynchronous and made a chained promise.
```Typescript
return await axios.getAll(/*EndPoint*/).catch( error => {return error.response? error.response:null})
``` 

Or use **_Try / Catch Pattern_** and catch the error synchronous and also made a chained promise.
``` Typescript 
try {
  return await axios.get(/*Endpoint*/);
} catch (error) {
  return error.response? error.response:null     
}
```

I found that using try/catch pattern is slightly faster than **.catch**. Also, both are in the Axios documentation.

FiveRockingFingers/FRF.Web/ClientApp/src/services/ArtifactService.ts
```diff
@@ -0,0 +1,57 @@
...
+
+const ARTIFACTS_URL = `${BASE_URL}api/Artifacts/`;
+
+class ArtifactService {
+
+    static getAll = async () => {
+        try {
+            return await axios.get(`${ARTIFACTS_URL}GetAll`);
+        } catch (error) {
+            return error.response ? error.response : null;
```
matiasbeckerle, line 12: Maybe you can replace all the nulls with some kind of custom defined message for yourselves? Remember these errors are mostly for developers so maybe you can declare a string constant saying something like... "Unknown error while calling API". That way if you have error.response ok, cool, otherwise you return that other string.

matiasbeckerle (approved):

FiveRockingFingers/FRF.Web/ClientApp/src/services/ArtifactService.ts
gonzalorom: Okay. I've changed all the null for error.message. If the request was made and the server responded with a status code that falls out of the range of 2xx trigger the `error.response`. If not, something happened in setting up the request that triggered an Error or the request was made, but no response was received, now we send an `error.message`.

PR #70, opened 2021-01-11
[FIVE-266] (ArtifactsController) Modificación controller artefactos para trabajar con nuevas respuestas del servicio

martinmena:
## Cambios realizados
- Se modificaron los métodos del controller de artefactos para adaptarlos a las nuevas respuestas del servicio. También se modificaron los tests de la misma manera.

## Nota importante
Solo se los modificó para que puedan seguir trabajando de la misma manera que antes. En otro ticket se incluirán los cambios necesario para aprovechar las modificaciones sobre las respuestas de los servicios. La idea de este cambio es que simplemente permita seguir funcionando todo.

### DIFF
--- a/FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
+++ b/FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
@@ -28,7 +28,7 @@ namespace FRF.Web.Controllers
         {
             var artifacts = await _artifactsService.GetAll();
 
-            var artifactsDto = _mapper.Map<IEnumerable<ArtifactDTO>>(artifacts);
+            var artifactsDto = _mapper.Map<IEnumerable<ArtifactDTO>>(artifacts.Value);
 
             return Ok(artifactsDto);
         }
@@ -38,7 +38,7 @@ namespace FRF.Web.Controllers
         {
             var artifacts = await _artifactsService.GetAllByProjectId(projectId);
 
-            var artifactsDto = _mapper.Map<IEnumerable<ArtifactDTO>>(artifacts);
+            var artifactsDto = _mapper.Map<IEnumerable<ArtifactDTO>>(artifacts.Value);
 
             return Ok(artifactsDto);
         }
@@ -48,12 +48,12 @@ namespace FRF.Web.Controllers
         {
             var artifact = await _artifactsService.Get(id);
 
-            if (artifact == null)
+            if (artifact.Success == false)
             {
                 return NotFound();
             }
 
-            var artifactDto = _mapper.Map<ArtifactDTO>(artifact);
+            var artifactDto = _mapper.Map<ArtifactDTO>(artifact.Value);
 
             return Ok(artifactDto);
         }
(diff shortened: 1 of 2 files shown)


FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
```diff
@@ -48,12 +48,12 @@ public async Task<IActionResult> GetAsync(int id)
         {
             var artifact = await _artifactsService.Get(id);
 
-            if (artifact == null)
+            if (artifact.Success == false)
```
matiasbeckerle, line 51: Or... !artifact.Success

FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
```diff
@@ -73,14 +74,15 @@ public async Task<IActionResult> UpdateAsync(int id, ArtifactUpsertDTO artifactD
         {
             var artifact = await _artifactsService.Get(id);
 
-            if (artifact == null)
+            if (artifact.Value == null)
```
matiasbeckerle, line 77: Question: when you can't find the resources on the service layer, you're going to return success or not?

martinmena: If it's a get, and you're searching for a specific artifact (in this case), it's returning an error because it doesn't exist. On the other hand, if you're searching for the artifacts associated to a specific project, and that project has no artifacts, the response would be a success, but an empty list. Is that ok?

matiasbeckerle: Yup, makes sense. Then in this particular case you should check for...

```csharp
if (!artifact.Successs)
{
  return NotFound();
}
```

FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
```diff
@@ -63,7 +63,8 @@ public async Task<IActionResult> SaveAsync(ArtifactUpsertDTO artifactDto)
         {
             var artifact = _mapper.Map<FRF.Core.Models.Artifact>(artifactDto);
 
-            var artifactCreated = _mapper.Map<ArtifactDTO>(await _artifactsService.Save(artifact));
+            var response = await _artifactsService.Save(artifact);
+            var artifactCreated = _mapper.Map<ArtifactDTO>(response.Value);
 
             return Ok(artifactCreated);
```
gonzalorom, line 69: Here I think we should return something if the Artifact couldn't be created. Perhaps some BadRequest(artifact.Error);

martinmena: Totally, but remember that the idea of this PR was only to adapt the controller methods to work with the new service response, so it can still work like it did before. There's going to be another ticket for the improvements in the methods that this new response allows us to make, like the one you mentioned.

gonzalorom: Awesome, thank you.

gonzalorom (approved):

FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
martinmena: Done!

matiasbeckerle (approved):

PR #71, opened 2021-01-12
[FIVE-289]Modificar rutas en llamadas a Axios de absolutas a relativas.

gonzalorom:
### Tarea realizada:
- Se modifican rutas absolutas a relativas en toda la capa cliente.

### DIFF
--- a/FiveRockingFingers/FRF.Web/ClientApp/src/Constants.tsx
+++ b/FiveRockingFingers/FRF.Web/ClientApp/src/Constants.tsx
@@ -1,4 +1,4 @@
 ﻿// Constants definitions for access on ClientApp
 
-export const BASE_URL = "https://localhost:44346/";
+export const BASE_URL = "/api/";
 export const PROVIDERS = ['AWS', 'Custom'];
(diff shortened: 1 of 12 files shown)


FiveRockingFingers/FRF.Web/ClientApp/src/Constants.tsx
```diff
@@ -1,4 +1,10 @@
 ﻿// Constants definitions for access on ClientApp
 
-export const BASE_URL = "https://localhost:44346/";
+export const PROJECTS_URL = "/api/Projects/";
+export const ARTIFACTS_URL = "/api/Artifacts/";
+export const CATEGORIES_URL = "/api/Categories/";
+export const COGNITO_SIGNIN_URL = "/api/SignIn";
+export const COGNITO_SIGNUP_URL = "/api/SignUp";
+export const AWS_ARTIFACTS_PROVIDER_URL = "/api/AwsArtifactsProvider/";
+export const USER_URL = "/api/User/";
```
matiasbeckerle, line 9: On one hand I want to recommend using the "base" naming in all these constants. Like `PROJECTS_BASE_URL` for example. On the other hand I'm not very in love with all these constants. I feel like here there should be only 1 URL similar to what you had already: `export const BASE_URL = "/api"` and then add more on each service, if you need. For example `BASE_USERS_URL = $'{BASE_URL}/Users'` which then you use on that service like...

```
axios.get($'{BASE_USERS_URL}/id...
```

gonzalorom: Done.

matiasbeckerle (approved):

PR #76, opened 2021-01-18
[FIVE-266] (ProjectsController) Modificación del controller de categorías para trabajar con nuevas respuestas del servicio

martinmena:
## Cambios realizados
- Se modificaron los métodos del controller de proyectos para adaptarlos a las nuevas respuestas del servicio. También se modificaron los tests de la misma manera.
- Se eliminaron los tests "Save_WhenProjectUpsertDTOIsInvalid_ReturnsBadRequest" y "Update_WhenGivenIdIsDifferentFromProjectId_BadRequest". El primero porque el chequeo de DTO se hace fuera del controller, y el segundo porque en algunos cambios de tickets anteriores se eliminó el ID en el ProjectUpsertDTO, por lo que no hay 2 Ids para comparar.
- Se crearon los tests "Update_WhenProjectDoesntExist_BadRequest" y "Update_WhenGivenEmptyUserList_BadRequest".
- Modificado el chequeo de status en ProjectsList al eliminar un proyecto (pasó de 204 a 200)

### DIFF
--- a/FiveRockingFingers/FRF.Web/ClientApp/src/components/ManageProjectsComponents/ConfirmationDialog.tsx
+++ b/FiveRockingFingers/FRF.Web/ClientApp/src/components/ManageProjectsComponents/ConfirmationDialog.tsx
@@ -15,7 +15,7 @@ export default function ConfirmationDialog(props: { keepMounted: boolean, open: 
             props.resetView(project.id);
             try {
                 const response = await ProjectService.delete(project.id.toString());
-                if (response.status === 204) {
+                if (response.status === 200) {
                     props.openSnackbar({ message: "Se eliminó el proyecto con éxito", severity: "success" });
                     updateProjects();
                 } else if (response.status === 404) {
--- a/FiveRockingFingers/FRF.Web/ClientApp/src/components/ManageProjectsComponents/ConfirmationDialog.tsx
+++ b/FiveRockingFingers/FRF.Web/ClientApp/src/components/ManageProjectsComponents/ConfirmationDialog.tsx
@@ -15,7 +15,7 @@ export default function ConfirmationDialog(props: { keepMounted: boolean, open: 
             props.resetView(project.id);
             try {
                 const response = await ProjectService.delete(project.id.toString());
-                if (response.status === 200) {
+                if (response.status === 204) {
                     props.openSnackbar({ message: "Se eliminó el proyecto con éxito", severity: "success" });
                     updateProjects();
                 } else if (response.status === 404) {
--- a/FiveRockingFingers/FRF.Web/Controllers/ProjectsController.cs
+++ b/FiveRockingFingers/FRF.Web/Controllers/ProjectsController.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using AutoMapper;
+﻿using AutoMapper;
 using FRF.Core.Models;
 using FRF.Core.Services;
 using FRF.Web.Dtos;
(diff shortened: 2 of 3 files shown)


FiveRockingFingers/FRF.Web/ClientApp/src/components/ManageProjectsComponents/ConfirmationDialog.tsx
```diff
@@ -15,7 +15,7 @@ export default function ConfirmationDialog(props: { keepMounted: boolean, open:
             props.resetView(project.id);
             try {
                 const response = await ProjectService.delete(project.id.toString());
-                if (response.status === 204) {
+                if (response.status === 200) {
```
matiasbeckerle, line 18: Just in case, a 204 response is acceptable in this case too.

FiveRockingFingers/FRF.Web/Controllers/ProjectsController.cs
```diff
@@ -59,39 +58,39 @@ public async Task<IActionResult> SaveAsync(ProjectUpsertDTO projectDto)
...
             var project = await _projectService.GetAsync(id);
-            if (project == null) return NotFound();
+            if (!project.Success) return NotFound();
 
             var isDeleted = await _projectService.DeleteAsync(id);
-            if (!isDeleted) return NotFound();
+            if (!isDeleted.Success) return NotFound();
 
-            return NoContent();
+            return Ok(isDeleted.Value);
```
matiasbeckerle, line 93: Just in case, a 204 response is acceptable in this case too.

test/FRF.Web.Tests/Controllers/ProjectsControllerTests.cs
```diff
@@ -200,37 +204,48 @@ public async Task Save_ReturnsOk()
...
             var project = CreateProject(projectId);
             var projectUpsertDTO = CreateProjectUpsertDto(project);
 
+            _projectsService
+                .Setup(mock => mock.GetAsync(It.IsAny<int>()))
+                .ReturnsAsync(new ServiceResponse<Project>(new Error(ErrorCodes.ProjectNotExists, "Error message")));
+
             // Act
-            var result = await _classUnderTest.SaveAsync(projectUpsertDTO);
+            var result = await _classUnderTest.UpdateAsync(project.Id + 1, projectUpsertDTO);
```
matiasbeckerle, line 220: I prefer to use 0 as ID in these scenarios (but it's just me being paranoid).

matiasbeckerle (approved):

FiveRockingFingers/FRF.Web/Controllers/ProjectsController.cs
martinmena: Isn't a little confusing returning the deleted project inside a 'No content' response? That's the only reason I thought 200 was better.

matiasbeckerle: Hehehe no -- I mean it's ok to return nothing in a NoContent response. Since you're deleting a resource kinda makes sense. I wasn't thinking in returning data in a 204 response, that's illegal, yes. You return something in a 200 or nothing with a 204.

martinmena: And between the two options, which one would you choose? I'm guessing it depends on the situation, but in this case, and based on what you said, I think the 204 may be better since the deleted project is not being used for anything right now, and there are no plans to use it in the future as far as I know. But in that case, what about the service? If I change this response to a 204, does that mean that I should change the service also, so it doesn't return the deleted project?

matiasbeckerle: As you inferred I'm more inclined towards a 204 response. Regarding the service I believe what you have is OK, let it return the object. Remember that APIs are just 1 type of presentation. Services instead, can be used elsewhere. Imagine you have some sort of background job server that needs to make use of services. Maybe that needs the value that is returned by the service.

martinmena: Sounds reasonable, I'll make the changes and keep it in mind for the next one. Thanks for the explanation!

gperelstein (approved):

PR #77, opened 2021-01-19
[Five-121] Crear pagina de relaciones entre artefactos

gperelstein:
Cambios realizados

-Se creó un XMLHelper para mapear las settings del XML con una lista
-Se agregaron los DTOs ArtifactsRelationInsertDTO y ArtifactsRelationUpdateDTO
-Se agregó modificó el mapeo entre Artifact y ArtifactDTO para que se utilice el XMLHelper en el campo de las settings
-Se agregó el código de Error 12 RelationNotExist
-Se agregó la interfaz ArtifactRelation
-Se modificó el DTO ArtifactDTO para que las Settings sea un diccionario en lugar de XElement


FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
```diff
@@ -11,5 +11,6 @@ public static class ErrorCodes
         // Relation errors
         public const int RelationAlreadyExisted = 10;
         public const int RelationNotValid = 11;
+        public const int RelationNotExist = 12;
```
matiasbeckerle, line 14: NotExist**s**

FiveRockingFingers/FRF.Web.Dtos/Artifacts/XMLHelper.cs
```diff
@@ -0,0 +1,45 @@
+﻿using AutoMapper;
+using FRF.Core.Models;
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Xml.Linq;
+
+namespace FRF.Web.Dtos.Artifacts
+{
+    public class XMLHelper : IValueResolver<Artifact, ArtifactDTO, Dictionary<string, string>>
```
matiasbeckerle, line 10: Maybe `XmlResolver`?

matiasbeckerle (approved):

FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
gperelstein: Done!

FiveRockingFingers/FRF.Web.Dtos/Artifacts/XMLHelper.cs
gperelstein: Yes, I think it's a better name. Done!

FiveRockingFingers/FRF.Web.Dtos/Artifacts/ArtifactsRelationUpdateDTO.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.ComponentModel.DataAnnotations;
+
+namespace FRF.Web.Dtos.Artifacts
+{
+    public class ArtifactsRelationUpdateDTO
+    {
+        [Required]
+        public Guid? Id { get; set; }
```
martinmena, line 9: Does it make sense to have a nullable field that's also required?

martinmena (approved):

FiveRockingFingers/FRF.Web.Dtos/Artifacts/ArtifactsRelationUpdateDTO.cs
gonzalorom: Yes, that is because GUID can not be null, because it is a struct and the default value is "00000000-0000-0000-0000-000000000000". Even if you use **[Required(AllowEmptyStrings = false)]**.
Dataannotation "**Required**" do something like `isValid: if value != null` so here (what I think and for sure there are more solutions) I can see two workarounds to avoid an empty GUID from the request:
1. Create a Custom Validation for "Required.GUID" for example `isValid: if(UserID != default(GUID))`
2. The itchy one:  first make GUID nullable by using the **?** notation and then, because now UserId is null, use the annotation "**Required**".

matiasbeckerle: It's tricky but for now I would recommend keeping this approach.

martinmena: Awesome, thanks for the explanation @gonzalorom!

PR #78, opened 2021-01-19
[FIVE-121 DAL] Crear pagina de relaciones entre artefactos

gonzalorom:
### Tareas realizadas:
- Se crearon métodos GetAllRelationsByProjectIdAsync, GetRelationsAsync, UpdateRelationAsync y DeleteRelationAsync en ArtifactsService.
- Se crearon Unit Test para los nuevos métodos de ArtifactsService.

### DIFF
--- a/FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
+++ b/FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
@@ -22,7 +22,7 @@ namespace FRF.Core.Services
             _mapper = mapper;
         }
 
-        private async Task<bool> IsAnyArtifactExcept (IList<ArtifactsRelation> artifactsRelations)
+        private async Task<bool> DoArtifactsExist (IList<ArtifactsRelation> artifactsRelations)
         {
             var dbArtifactsId = await _dataContext.Artifacts.Select(a => a.Id).ToListAsync();
             var artifactsRelationIds = artifactsRelations
@@ -32,13 +32,19 @@ namespace FRF.Core.Services
             return artifactsRelationIds.Except(dbArtifactsId).Any();
         }
 
-        private bool IsAnyRelationRepeated(IList<EntityModels.ArtifactsRelation> dbArtifactRelations, IList<ArtifactsRelation> artifactsRelations)
+        private bool IsAnyRelationRepeated(IList<EntityModels.ArtifactsRelation> dbArtifactRelations,
+            IList<ArtifactsRelation> artifactsRelations)
         {
-           
             return artifactsRelations.Any(ar => dbArtifactRelations.Any(dbAr =>
-                dbAr.Artifact1Id == ar.Artifact1Id && dbAr.Artifact2Id == ar.Artifact2Id &&
-                dbAr.Artifact1Property.Equals(ar.Artifact1Property, StringComparison.InvariantCultureIgnoreCase) &&
-                dbAr.Artifact2Property.Equals(ar.Artifact2Property, StringComparison.InvariantCultureIgnoreCase) || dbAr.Id == ar.Id));
+                (dbAr.Artifact1Id == ar.Artifact1Id && dbAr.Artifact2Id == ar.Artifact2Id &&
+                 dbAr.Artifact1Property.Equals(ar.Artifact1Property, StringComparison.InvariantCultureIgnoreCase) &&
+                 dbAr.Artifact2Property.Equals(ar.Artifact2Property, StringComparison.InvariantCultureIgnoreCase) &&
+                 dbAr.RelationTypeId == ar.RelationTypeId)
+                ||
+                (dbAr.Artifact1Id == ar.Artifact2Id && dbAr.Artifact2Id == ar.Artifact1Id &&
+                 dbAr.Artifact1Property.Equals(ar.Artifact2Property, StringComparison.InvariantCultureIgnoreCase) &&
+                 dbAr.Artifact2Property.Equals(ar.Artifact1Property, StringComparison.InvariantCultureIgnoreCase) &&
+                 dbAr.RelationTypeId == ar.RelationTypeId)));
         }
 
         public async Task<ServiceResponse<List<Artifact>>> GetAll()
@@ -174,16 +180,18 @@ namespace FRF.Core.Services
 
         public async Task<ServiceResponse<IList<ArtifactsRelation>>> SetRelationAsync(IList<ArtifactsRelation> artifactRelations)
         {
-            var isAnyArtifactExcept = await IsAnyArtifactExcept(artifactRelations);
-            if (isAnyArtifactExcept) return new ServiceResponse<IList<ArtifactsRelation>>(new Error(ErrorCodes.RelationNotValid, "At least one of the artifact Ids provided doesn't exist"));
+            var artifactsExist = await DoArtifactsExist(artifactRelations);
+            if (artifactsExist) return new ServiceResponse<IList<ArtifactsRelation>>(new Error(ErrorCodes.RelationNotValid, "At least one of the artifact Ids provided doesn't exist"));
 
             var dbArtifactRelations = await _dataContext.ArtifactsRelation.Where(ar =>
                     ar.Artifact1Id == artifactRelations[0].Artifact1Id ||
-                    ar.Artifact2Id == artifactRelations[0].Artifact2Id)
+                    ar.Artifact2Id == artifactRelations[0].Artifact2Id ||
+                    ar.Artifact2Id == artifactRelations[0].Artifact1Id ||
+                    ar.Artifact1Id == artifactRelations[0].Artifact2Id)
                 .ToListAsync();
 
-            var isAnyArtifactRepeated = IsAnyRelationRepeated(dbArtifactRelations, artifactRelations);
-            if (isAnyArtifactRepeated)
+            var relationsRepeated = IsAnyRelationRepeated(dbArtifactRelations, artifactRelations);
+            if (relationsRepeated)
                 return new ServiceResponse<IList<ArtifactsRelation>>(new Error(ErrorCodes.RelationAlreadyExisted, "At least one of the relations already existed"));
 
             var resultArtifactRelations = _mapper.Map<IList<EntityModels.ArtifactsRelation>>(artifactRelations);
@@ -193,14 +201,19 @@ namespace FRF.Core.Services
             return new ServiceResponse<IList<ArtifactsRelation>>(_mapper.Map<IList<ArtifactsRelation>>(resultArtifactRelations));
         }
 
-        public async Task<IList<ArtifactsRelation>> GetRelationsAsync(int artifactId)
+        public async Task<ServiceResponse<IList<ArtifactsRelation>>> GetAllRelationsOfAnArtifactAsync(int artifactId)
         {
-            var artifactsRelations = await _dataContext.ArtifactsRelation
-                .Include(ar => ar.Artifact1)
+            var result = await _dataContext.ArtifactsRelation.Include(ar => ar.Artifact1)
                 .Include(ar => ar.Artifact2)
-                .Where(ar => ar.Artifact1Id == artifactId || ar.Artifact2Id == artifactId).ToListAsync();
-            var resultArtifactRelations = _mapper.Map<List<ArtifactsRelation>>(artifactsRelations);
-            return resultArtifactRelations;
+                .Where(ar => ar.Artifact1Id == artifactId || ar.Artifact2Id == artifactId)
+                .ToListAsync();
+
+            if (!result.Any())
+                return new ServiceResponse<IList<ArtifactsRelation>>(new Error(ErrorCodes.RelationNotExist,
+                    $"There is no relations for the artifact with Id = {artifactId}"));
+
+            var resultArtifactRelations = _mapper.Map<List<ArtifactsRelation>>(result);
+            return new ServiceResponse<IList<ArtifactsRelation>>(resultArtifactRelations);
         }
 
         public async Task<ServiceResponse<IList<ArtifactsRelation>>> GetAllRelationsByProjectIdAsync(int projectId)
@@ -212,7 +225,7 @@ namespace FRF.Core.Services
             }
 
             var artifactsRelations = await _dataContext.ArtifactsRelation
-                .Where(ar => ar.Artifact1.ProjectId == projectId)
+                .Where(ar => ar.Artifact1.ProjectId == projectId || ar.Artifact2.ProjectId == projectId)
                 .Include(ar => ar.Artifact1)
                 .Include(ar => ar.Artifact2)
                 .ToListAsync();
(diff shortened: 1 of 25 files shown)


FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -23,6 +22,25 @@ public ArtifactsService(DataAccessContext dataContext, IMapper mapper)
...
+            return artifactsRelationIds.Except(dbArtifactsId).Any();
+        }
+
+        private bool IsAnyRelationRepeated(IList<EntityModels.ArtifactsRelation> dbArtifactRelations, IList<ArtifactsRelation> artifactsRelations)
+        {
+           
+            return artifactsRelations.Any(ar => dbArtifactRelations.Any(dbAr =>
+                dbAr.Artifact1Id == ar.Artifact1Id && dbAr.Artifact2Id == ar.Artifact2Id &&
+                dbAr.Artifact1Property.Equals(ar.Artifact1Property, StringComparison.InvariantCultureIgnoreCase) &&
+                dbAr.Artifact2Property.Equals(ar.Artifact2Property, StringComparison.InvariantCultureIgnoreCase) || dbAr.Id == ar.Id));
```
matiasbeckerle, line 41: You have ANDS and one OR at the end without much use of parenthesis. Are you sure these conditions are OK?

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -156,38 +174,102 @@ public async Task<ServiceResponse<Artifact>> Delete(int id)
...
-            var isAnyArtifactRepeated = artifactRelations
-                .Any(ar => dbArtifactRelations
-                    .Any(dbAr => 
-                        dbAr.Artifact1Id == ar.Artifact1Id
-                        && dbAr.Artifact2Id == ar.Artifact2Id
-                        && dbAr.Artifact1Property.Equals(ar.Artifact1Property, StringComparison.InvariantCultureIgnoreCase)
-                        && dbAr.Artifact2Property.Equals(ar.Artifact2Property, StringComparison.InvariantCultureIgnoreCase)
-                        )
-                );
+            var isAnyArtifactExcept = await IsAnyArtifactExcept(artifactRelations);
```
matiasbeckerle, line 177: Not sure I understand the meaning of this method and variable name. What's the purpose?

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -156,38 +174,102 @@ public async Task<ServiceResponse<Artifact>> Delete(int id)
...
-                await _dataContext.ArtifactsRelation.AddAsync(mappedArtifactRelation);
-                resultArtifactRelations.Add(artifactRelation);
+            var resultArtifactRelations = _mapper.Map<IList<EntityModels.ArtifactsRelation>>(artifactRelations);
+            await _dataContext.ArtifactsRelation.AddRangeAsync(resultArtifactRelations);
+            await _dataContext.SaveChangesAsync();
+
+            return new ServiceResponse<IList<ArtifactsRelation>>(_mapper.Map<IList<ArtifactsRelation>>(resultArtifactRelations));
+        }
+
+        public async Task<IList<ArtifactsRelation>> GetRelationsAsync(int artifactId)
```
matiasbeckerle, line 196: I'm having a hard time trying to figure it out what's happening in this service. Is there any chances to use semantic method names instead of technical names? For example I feel this method is named based on what happens at the database level (a relationship between two or the same entity). But what it really means? Maybe this method can't be renamed but maybe some of the others?

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -156,38 +174,102 @@ public async Task<ServiceResponse<Artifact>> Delete(int id)
...
+                            _mapper.Map<EntityModels.ArtifactsRelation>(relationNew));
+                    }
+
+                    relationOriginal.Artifact1Id = relationNew.Artifact1Id;
+                    relationOriginal.Artifact2Id = relationNew.Artifact2Id;
+                    relationOriginal.Artifact1Property = relationNew.Artifact1Property;
+                    relationOriginal.Artifact2Property = relationNew.Artifact2Property;
+                    relationOriginal.RelationTypeId = relationNew.RelationTypeId;
+                }
             }
```
matiasbeckerle, line 268: What's happening in this foreach?

matiasbeckerle (approved):
All looks good but code is hard to read/understand. It feels like everything is named based on what you do at database level but that doesn't have an abstraction or semantic naming. Maybe you can consider some renames or grouping some functionality in new methods with a semantic name?

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gonzalorom: It is a good practice in C# overwrite `Equals()` and `GetHashCode()` on the class? 
In that case we can avoid doing all the ANDS. Because here we compare if is any Artifact Relation **equal** (ANDS) or if the Artifact Relation ID already exist.

martinmena, line 41: What happens if I make the same relation, but the other way around changing the RelationTypeId so it keeps the same logic?

Relation 1:
- ArtifactID1 = 10
- ArtifactID2 = 11
- Artifact1Property = "anything1"
- Artifact2Property = "anything2"
- RelationTypeId = 1 (I don't remember which one is which, but let's say 1 is from the Artifact1 to Artifact2)

Relation 2:
- ArtifactID1 = 11
- ArtifactID2 = 10
- Artifact1Property = "anything2"
- Artifact2Property = "anything1"
- RelationTypeId = 2 (I don't remember which one is which, but let's say 2 is from the Artifact2 to Artifact1)

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gonzalorom: Here the meaning of this method is to answer the question: **Is there an artifact id that doesn't exist?** . Do you have any suggestion for the name?

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -156,38 +174,102 @@ public async Task<ServiceResponse<Artifact>> Delete(int id)
...
-                        && dbAr.Artifact2Property.Equals(ar.Artifact2Property, StringComparison.InvariantCultureIgnoreCase)
-                        )
-                );
+            var isAnyArtifactExcept = await IsAnyArtifactExcept(artifactRelations);
+            if (isAnyArtifactExcept) return new ServiceResponse<IList<ArtifactsRelation>>(new Error(ErrorCodes.RelationNotValid, "At least one of the artifact Ids provided doesn't exist"));
+
+            var dbArtifactRelations = await _dataContext.ArtifactsRelation.Where(ar =>
+                    ar.Artifact1Id == artifactRelations[0].Artifact1Id ||
+                    ar.Artifact2Id == artifactRelations[0].Artifact2Id)
+                .ToListAsync();
```
martinmena, line 183: Similar comment as the previous one, you're getting every relation that has the Artifact1ID in the first position or the Artifact2ID in the second one, but they could also be the other way around.
Also, I'm not quite sure if you're checking for every artifact relation, or just the ones that have the Ids from the first relation in the array.
For example, let's say artifactRelations has this ids in each position: [[0, 1], [0, 1], [1, 0], [2, 3]].
If that's the case you would be checking only the relations that have a 0 in Artifact1ID or a 1 on Artifact2ID, but not the other way around, or the last relation if I'm not mistaken ([2, 3]), but I don't know if this is a possible case, or at this point artifactRelations only has relations from Id=0 to Id=1

FiveRockingFingers/FRF.Core/Services/IArtifactsService.cs
```diff
@@ -14,5 +15,10 @@ public interface IArtifactsService
         Task<ServiceResponse<Artifact>> Delete(int id);
         Task<ServiceResponse<Artifact>> Save(Artifact artifact);
         Task<ServiceResponse<IList<ArtifactsRelation>>> SetRelationAsync(IList<ArtifactsRelation> artifactRelations);
+        Task<IList<ArtifactsRelation>> GetRelationsAsync(int artifactId);
```
martinmena, line 18: Why does this method return an "IList<ArtifactsRelations>" instead of "ServiceResponse<IList<ArtifactsRelations>>"?

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -156,38 +174,102 @@ public async Task<ServiceResponse<Artifact>> Delete(int id)
...
+
+        public async Task<IList<ArtifactsRelation>> GetRelationsAsync(int artifactId)
+        {
+            var artifactsRelations = await _dataContext.ArtifactsRelation
+                .Include(ar => ar.Artifact1)
+                .Include(ar => ar.Artifact2)
+                .Where(ar => ar.Artifact1Id == artifactId || ar.Artifact2Id == artifactId).ToListAsync();
+            var resultArtifactRelations = _mapper.Map<List<ArtifactsRelation>>(artifactsRelations);
+            return resultArtifactRelations;
+        }
```
martinmena, line 204: You're getting every relation that has the Artifact1ID in the first position or the Artifact2ID in the second one, but they could also be the other way around.

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -156,38 +174,102 @@ public async Task<ServiceResponse<Artifact>> Delete(int id)
...
+            }
+
+            var artifactsRelations = await _dataContext.ArtifactsRelation
+                .Where(ar => ar.Artifact1.ProjectId == projectId)
+                .Include(ar => ar.Artifact1)
+                .Include(ar => ar.Artifact2)
+                .ToListAsync();
+            var resultArtifactRelations = _mapper.Map<List<ArtifactsRelation>>(artifactsRelations);
+            return new ServiceResponse<IList<ArtifactsRelation>>(resultArtifactRelations);
+        }
```
martinmena, line 221: You're getting every relation that has the Artifact1ID in the first position or the Artifact2ID in the second one, but they could also be the other way around.

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gperelstein: You shouldn't be able to do the same relation with the same artifacts and the same properties only changing the RelationTypeId. The RelationTypeId could be 0(from right to left), 1(from left to right), and 2(double way). If you have one relation from right to left and you want to add the same relation from left to right you should  change it to be double way

FiveRockingFingers/FRF.Core/Services/IArtifactsService.cs
gperelstein: Sorry, my bad! I forget about this one

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
matiasbeckerle: Besides that I was more interested in the OR clause at the end. It's weird but I believe you have something like:

```if (a && b && c && d && e || f)```

Notice there are no parenthesis separating the conditionals. It's valid code but I wonder if it's what you intended to write.

Comment (matiasbeckerle) [reply FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs:41]: I feel you intended to write:

`if ((a && b && c && d && e) || f)`

You know what I mean?

File: FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs (hunk 2 above)
Comment (matiasbeckerle) [reply FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs:177]: The _except_ at the end is confusing to me. Maybe you intended something like...

```
var artifactsExist = await DoArtifactsExist(artifactRelations);
if (!artifactsExist) return new ServiceResponse<IList<ArtifactsRelation>>(new Error(ErrorCodes.RelationNotValid, "One or more artifact Ids doesn't exist"));
```

File: FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs (hunk 1 above)
Comment (gperelstein) [reply FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs:41]: Sorry, reading better the example you gave I understand what you are trying to say. We are only checking if the artifact1 is column1 and artifact2 is column2 but not the other way around. We should check this

File: FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs (hunk 7 above)
Comment (gonzalorom) [reply FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs:204]: Here we need to get all the relations of only one artifact, so for that reason we are getting the ones that has the given `artifactId` in `Artiact1Id` or in `Artifact2Id`.

File: FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs (hunk 1 above)
Comment (gonzalorom) [reply FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs:41]: Yes I know what you mean, and you are right. Now I took out the OR in this method because we don't need it. Thank you for the explanation.

File: FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs (hunk 2 above)
Comment (gonzalorom) [reply FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs:177]: That way looks cleaner. Done.

File: FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs (hunk 7 above)
Comment (martinmena) [reply FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs:204]: Ohh my bad, you're right.

File: FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs [hunk 9]
```diff
@@ -23,6 +22,31 @@ public ArtifactsService(DataAccessContext dataContext, IMapper mapper)
...
+                (dbAr.Artifact1Id == ar.Artifact1Id && dbAr.Artifact2Id == ar.Artifact2Id &&
+                 dbAr.Artifact1Property.Equals(ar.Artifact1Property, StringComparison.InvariantCultureIgnoreCase) &&
+                 dbAr.Artifact2Property.Equals(ar.Artifact2Property, StringComparison.InvariantCultureIgnoreCase) &&
+                 dbAr.RelationTypeId == ar.RelationTypeId)
+                ||
+                (dbAr.Artifact1Id == ar.Artifact2Id && dbAr.Artifact2Id == ar.Artifact1Id &&
+                 dbAr.Artifact1Property.Equals(ar.Artifact2Property, StringComparison.InvariantCultureIgnoreCase) &&
+                 dbAr.Artifact2Property.Equals(ar.Artifact1Property, StringComparison.InvariantCultureIgnoreCase) &&
+                 dbAr.RelationTypeId == ar.RelationTypeId)));
+        }
```
martinmena, line 48: Sorry for insisting on the same, but I don't think this solves the issue. I think the same example as before would be possible, even tho it shouldn't, since it had different RelationTypeId's. In this case, you're still allowing the same relation twice, and another one in both directions.

| Row | Artifact1ID | Artifact1Property | Artifact2Id | Artifact2Property | RelationTypeId |
|------|-------------|-------------------|-------------|-------------------|----------------|
| (1) | 1           | Prop1             | 2           | Prop2             | 0 (-->)        |
| (2) | 2           | Prop2             | 1           | Prop1             | 1 (<--)        |
| (3) | 1           | Prop1             | 2           | Prop2             | 1 (<--)        |
| (4) | 2           | Prop2             | 1           | Prop1             | 0 (-->)        |
| (5) | 1           | Prop1             | 2           | Prop2             | 2 (<->)        |

In this case, the first two rows (1 and 2) are the same relation expressed differently, the next two (3 and 4), another relation expressed differently, and the last row represents both of the relations, so basically all the rows above, and I think as it is right now, all of this is possible.

Maybe the best idea is to avoid comparing the RelationType, and just checking if there's already a relation between the properties from one artifact to the other. That way you're only allowing one row, no matter the direction, and still allowing every type of relation. (@gperelstein commented something like this)

gonzalorom: OMG! Good catch I forgot to delete the RelationType once I was done with the testing.

PR #79, opened 2021-01-19
[Five-121-Controller] Crear pagina de relaciones entre artefactos

gperelstein:
Cambios realizados

 - Se crearon endpoint correspondientes a GetRelationAsync, GetAllRelationsByProjectIdAsync, UpdateRelationsAsync y DeleteRelationAsync en ArtifactsController.
- Se crearon Unit Test correspondientes a los nuevos endpoint de ArtifactsController.

### DIFF
--- a/FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
+++ b/FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
@@ -116,7 +116,7 @@ namespace FRF.Web.Controllers
         }
 
         [HttpGet]
-        [Route("api/[controller]/{artifactId}/[action]")]
+        [Route("~/api/[controller]/{artifactId}/[action]")]
         public async Task<IActionResult> GetRelationsAsync(int artifactId)
         {
             var result = await _artifactsService.GetAllRelationsOfAnArtifactAsync(artifactId);
@@ -150,7 +150,7 @@ namespace FRF.Web.Controllers
         }
 
         [HttpPut]
-        [Route("api/[controller]/{artifactId}/[action]")]
+        [Route("~/api/[controller]/{artifactId}/[action]")]
         public async Task<IActionResult> UpdateRelationsAsync(int artifactId,
             IList<ArtifactsRelationUpdateDTO> artifactRelationUpdatedList)
         {
--- a/FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
+++ b/FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
@@ -1,4 +1,5 @@
-﻿using AutoMapper;
+﻿using System;
+using AutoMapper;
 using FRF.Core.Models;
 using FRF.Core.Services;
 using FRF.Web.Dtos.Artifacts;
@@ -103,10 +104,56 @@ namespace FRF.Web.Controllers
         }
 
         [HttpPost]
-        public async Task<IActionResult> SetRelationAsync(IList<ArtifactsRelationDTO> artifactRelationList)
+        public async Task<IActionResult> SetRelationAsync(IList<ArtifactsRelationInsertDTO> artifactRelationList)
         {
             var artifactsRelations = _mapper.Map<IList<ArtifactsRelation>>(artifactRelationList);
             var result = await _artifactsService.SetRelationAsync(artifactsRelations);
+            if (!result.Success) return BadRequest();
+
+            var artifactsResult = _mapper.Map<IList<ArtifactsRelationDTO>>(result.Value);
+            return Ok(artifactsResult);
+        }
+
+        [HttpGet("{arifactId}")]
+        public async Task<IActionResult> GetRelationAsync(int artifactId)
+        {
+            var result = await _artifactsService.GetRelationsAsync(artifactId);
+
+            return Ok(_mapper.Map<IList<ArtifactsRelationDTO>>(result));
+        }
+
+        [HttpGet("{projectId}")]
+        public async Task<IActionResult> GetAllRelationsByProjectIdAsync(int projectId)
+        {
+            var result = await _artifactsService.GetAllRelationsByProjectIdAsync(projectId);
+
+             if (!result.Success) return BadRequest();
+
+             return Ok(_mapper.Map<IList<ArtifactsRelationDTO>>(result.Value));
+
+        }
+
+        [HttpDelete("{relationId}")]
+        public async Task<IActionResult> DeleteRelationAsync(Guid relationId)
+        {
+            var result = await _artifactsService.DeleteRelationAsync(relationId);
+            if (!result.Success)
+            {
+                return NotFound();
+            }
+
+            return NoContent();
+        }
+
+        [HttpPut("{artifactId}")]
+        public async Task<IActionResult> UpdateRelationsAsync(int artifactId,
+            IList<ArtifactsRelationUpdateDTO> artifactRelationUpdatedList)
+        {
+            var artifact = await _artifactsService.Get(artifactId);
+            if (!artifact.Success) return NotFound();
+
+            var artifactsRelationsList = _mapper.Map<IList<ArtifactsRelation>>(artifactRelationUpdatedList);
+            var result = await _artifactsService.UpdateRelationAsync(artifactId, artifactsRelationsList);
             if (!result.Success) return BadRequest();
 
             var artifactsResult = _mapper.Map<IList<ArtifactsRelationDTO>>(result.Value);
--- a/FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
+++ b/FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
@@ -115,8 +115,8 @@ namespace FRF.Web.Controllers
         }
 
         [HttpGet]
-        [Route("api/[controller]/{artifactId}/[action]")]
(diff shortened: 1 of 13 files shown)


FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
```diff
@@ -112,5 +113,51 @@ public async Task<IActionResult> SetRelationAsync(IList<ArtifactsRelationDTO> ar
             var artifactsResult = _mapper.Map<IList<ArtifactsRelationDTO>>(result.Value);
             return Ok(artifactsResult);
         }
+
+        [HttpGet("{artifactId}")]
```
matiasbeckerle, line 117: What's the final URL for this endpoint?

FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
```diff
@@ -112,5 +113,51 @@ public async Task<IActionResult> SetRelationAsync(IList<ArtifactsRelationDTO> ar
...
+            }
+
+            return NoContent();
+        }
+
+        [HttpPut("{artifactId}")]
+        public async Task<IActionResult> UpdateRelationsAsync(int artifactId,
+            IList<ArtifactsRelationUpdateDTO> artifactRelationUpdatedList)
+        {
+            var artifact = await _artifactsService.Get(artifactId);
```
matiasbeckerle, line 152: Maybe this could happen inside `_artifactsService.UpdateRelationAsync`

FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
gperelstein: The URL of this endpoint is for now is `api/Artifacts/GetRelation/:artifactId`. Maybe the artifact Id should be before GetRelation, right? Something like: `api/Artifacts/:artifactId/GetRelation`

matiasbeckerle: Ah ok it's what I suspected. To me it feels like the URLs in all these endpoints are not very restful or semantic. This is also related to what I was mentioning on the other PRs.

You will have to wrap your head around these concepts and think in some other URLs. For example, what does this endpoint mean? To obtain what? What's a relationship? For sure it should start with...

```
/api/artifacts/{artifactId}/...
```

gperelstein: I understand what you are saying. We didn't follow the bests practices to create restful apis. For example, we use a lot of verbs like GetRelation. Maybe a correct URL could be `api/artifacts/{artifactId}/relations`, right? The problem is that we didn't follow this practice in all of our endpoints (except for a few). Most of our URLs look something like `api/[controller]/[action]`. Maybe be could make a ticket to check all the URLs of the endpoints and make sure they follow the best practices. What do you think? For now I change this URL and the one corresponding to `UpdateRelationsAsync` method to be api/Artifacts/{artifactId}/GetRelations and api/Artifacts/{artifactId}/UpdateRelations

matiasbeckerle: To work on the URLs as a separate task it's fair and I'm glad you will consider thinking in RESTful. Using the verbs as part of the URL is a no-go (get, update, delete...).

Relations is a super broad word. What does it means?

gperelstein: The word relations it's referred to the relations between the properties of the artifacts

matiasbeckerle: Sorry to be that annoying but do you mind explaining a little bit, maybe using an example? I would like to better understand what's the meaning of _relations_ in this business.

gperelstein: No need to be sorry! Maybe I should be sorry for not explaining myself very well. In this ticket, we are trying to create relations between properties (or settings) of artifacts. For example:

Artifact1:
Settings={Name1: Valor1, Name2: Valor2}

Artifact2:
Settings={Name3: Valor3, Name4: Valor4}

We want to create relations between the settings, for example, Name1 -> Name3. We are going to use this in order to calculate the final price of the project. One setting could be the amount of information that we have to store in GB and the other setting could be the type of database that we are going to use to store the information. So knowing these two things we can calculate a price. I don't know if this time I was more clear

gonzalorom: For example this:
>### Artifact 1 Traffic Camera
>Property: Send Files
>Property Value: 1

>### Artifact 2 Database
> Property: Storage per day
> Property Value: 10

Artifact Relation:
<table>
  <tbody>
    <tr>
      <th align="center">Artifact</th>
      <th align="center">Direction</th>
      <th align="center">Artifact</th>
    </tr>
    <tr>
      <td align="right-side">
        <ul>
          <li>Artifact id 1: Traffic Camera</li>
          <li>Property: Send Files</li>
        </ul>
      </td>
      <td align="center">➡</td>
      <td align="right-side">
	  <ul>
          <li>Artifact id 2: Database</li>
          <li>Property: Storage per day</li>
        </ul>
		</td>
    </tr>
  </tbody>
</table>

FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
gperelstein: Done!

FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
matiasbeckerle: Thanks for the examples folks! Unfortunately I can't understand it yet hehe -- that's probably because the business of what you're building is completely foreign to me. Though, it looks like you're using database naming "relations" (table relationships) in order to give it a name to your business. So to give an example: what does the table above means?

I will give you an example of the point of view I'm trying to make. Let's imagine a videoclub. They rent movies. So they have a movies table. But they also have clients or customers. How do we represent the relation between the two? With a table containing the customer_id + movie_id. Do we call it MovieRelation? CustomerRelation? MovieCustomerRelation? or do we give it a better and business name like Rental(s)?

So on your example I feel like artifacts/id/relations isn't clear. Instead, following the videoclub you could even have two endpoints depending on your point of view:

```
/customers/{id}/rentals
/movies/{id}/rentals
```

Rental is the keyword. I'm trying to find rental in your explanations 😅.

matiasbeckerle (approved):
Besides code clarity and the endpoint routes I discussed in some of the comments, this looks safe to merge. I believe those other two things can be improved over time if you have them on mind.

FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
```diff
@@ -112,5 +114,53 @@ public async Task<IActionResult> SetRelationAsync(IList<ArtifactsRelationDTO> ar
...
+
+        [HttpPut]
+        [Route("api/[controller]/{artifactId}/[action]")]
+        public async Task<IActionResult> UpdateRelationsAsync(int artifactId,
+            IList<ArtifactsRelationUpdateDTO> artifactRelationUpdatedList)
+        {
+            var artifactsRelationsList = _mapper.Map<IList<ArtifactsRelation>>(artifactRelationUpdatedList);
+            var result = await _artifactsService.UpdateRelationAsync(artifactId, artifactsRelationsList);
+            if(!result.Success && result.Error.Code == ErrorCodes.ArtifactNotExists) return NotFound();
+            if (!result.Success && result.Error.Code == ErrorCodes.RelationNotValid) return BadRequest();
```
matiasbeckerle, line 160: Check this out folks: these two lines above represents what we have been thinking when we talked about the ServiceResponse class. Here, now, you can identify what happened and act accordingly with a notfound or badrequest. Good job!

martinmena (approved):
Just as a thought, every _"UpdateArtifactsRelationsAsync"_ test is basically the same, except for the service mock. In all of them you have the same input, but different output for some reason. I don't know how important that is, just wanted to mention it. Everything else looks great!

FiveRockingFingers/FRF.Web/Controllers/ArtifactsController.cs
matiasbeckerle: @gperelstein @gonzalorom maybe this comment went off the radar so I just want to make sure you saw it.

gonzalorom: Yes @matiasbeckerle we saw it. Yesterday we put this issue on the table with @egutbrod and the rest of the team. And we couldn't come up with a different synonym or reference to use as a name. I apologize for not being able to explain it in a more semantic way.

matiasbeckerle: Np -- just wanted to be sure you read it. It's ok to move forward and have this in mind in the future.

gperelstein: Yes, I read it too. The problem that we have is that artifacts could be a lot of different things. Could be a service of Amazon or something custom created for the user that could involve that amount of data that we are going to have to store or the number of requests that the app is going to manage. These artifacts are going to be related to price calculations but these relations are abstracts

PR #83, opened 2021-01-20
[FIVE-266] (SignUp) Incorporación de ServiceResponse en SignUpService y su controller

martinmena:
## Cambios realizados
- Se creó un nuevo código de error para los servicios (SignUpError = 22).
- Se modificó la interfaz del servicio de SignUp para que el método trabaje con objetos ServiceResponse.
- Se realizaron los cambios en el servicio de SignUp y en sus tests de acuerdo a los cambios en la interfaz.
- Se modificó el controller de SignUp para trabajar con las nuevas respuestas del servicio.

### DIFF
--- a/FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
+++ b/FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
@@ -13,6 +13,7 @@
         public const int RelationNotValid = 11;
 
         // Users
-        public const int SignUpError = 22;
+        public const int InvalidCredentials = 20;
+        public const int AuthenticationServerCurrentlyUnavailable = 21;
     }
 }
--- a/FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
+++ b/FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
@@ -11,5 +11,8 @@
         // Relation errors
         public const int RelationAlreadyExisted = 10;
         public const int RelationNotValid = 11;
+
+        // Users
+        public const int SignUpError = 22;
     }
 }
--- a/FiveRockingFingers/FRF.Core/Services/SignUpService.cs
+++ b/FiveRockingFingers/FRF.Core/Services/SignUpService.cs
@@ -39,7 +39,7 @@ namespace FRF.Core.Services
                 var response = await _userManager.CreateAsync(newCognitoUser, newUser.Password);
 
                 if (!response.Succeeded)
-                    return new ServiceResponse<string>(new Error(ErrorCodes.InvalidCredentials, ""));
+                    return new ServiceResponse<string>(new Error(ErrorCodes.InvalidCredentials, "There was an error with your email and/or password"));
 
                 //Auto validate the new user. They will not be included in the production build
                 await _cognitoIdentity.AdminUpdateUserAttributesAsync(new AdminUpdateUserAttributesRequest
@@ -60,17 +60,17 @@ namespace FRF.Core.Services
 
                 var token = await _signInManager.UserManager.FindByEmailAsync(newUser.Email);
                 if (token == null)
-                    return new ServiceResponse<string>(new Error(ErrorCodes.AuthenticationServerCurrentlyUnavailable, ""));
+                    return new ServiceResponse<string>(new Error(ErrorCodes.AuthenticationServerCurrentlyUnavailable, "There was an error with the Authentication service"));
 
                 var result =
                     await _signInManager.PasswordSignInAsync(token, newUser.Password, false, false);
                 return result.Succeeded
                     ? new ServiceResponse<string>(token.UserID)
-                    : new ServiceResponse<string>(new Error(ErrorCodes.AuthenticationServerCurrentlyUnavailable, ""));
+                    : new ServiceResponse<string>(new Error(ErrorCodes.AuthenticationServerCurrentlyUnavailable, "There was an error with the Authentication service"));
             }
             catch
             {
-                return new ServiceResponse<string>(new Error(ErrorCodes.AuthenticationServerCurrentlyUnavailable, ""));
+                return new ServiceResponse<string>(new Error(ErrorCodes.AuthenticationServerCurrentlyUnavailable, "There was an error with the Authentication service"));
 
             }
         }
--- a/FiveRockingFingers/FRF.Core/Services/SignUpService.cs
+++ b/FiveRockingFingers/FRF.Core/Services/SignUpService.cs
@@ -3,6 +3,7 @@ using Amazon.CognitoIdentityProvider;
 using Amazon.CognitoIdentityProvider.Model;
 using Amazon.Extensions.CognitoAuthentication;
 using FRF.Core.Models;
+using FRF.Core.Response;
 using Microsoft.AspNetCore.Identity;
 using System;
 using System.Collections.Generic;
@@ -26,7 +27,7 @@ namespace FRF.Core.Services
             _pool = pool;
         }
 
-        public async Task<Tuple<bool, string>> SignUpAsync(User newUser)
+        public async Task<ServiceResponse<string>> SignUpAsync(User newUser)
         {
             try
             {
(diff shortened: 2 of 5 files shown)


FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
```diff
@@ -11,5 +11,8 @@ public static class ErrorCodes
         // Relation errors
         public const int RelationAlreadyExisted = 10;
         public const int RelationNotValid = 11;
+
+        // Users
+        public const int SignUpError = 22;
```
matiasbeckerle, line 16: Remember that everything listed here is an error. What does this really means?

FiveRockingFingers/FRF.Core/Services/SignUpService.cs
```diff
@@ -38,7 +39,7 @@ public class SignUpService : ISignUpService
                 var response = await _userManager.CreateAsync(newCognitoUser, newUser.Password);
 
                 if (!response.Succeeded)
-                    return new Tuple<bool, string>(false, "");
+                    return new ServiceResponse<string>(new Error(ErrorCodes.SignUpError, ""));
```
gonzalorom, line 42: Here you will get a not succeeded response if the email already exist.

FiveRockingFingers/FRF.Core/Services/SignUpService.cs
```diff
@@ -58,13 +59,14 @@ public class SignUpService : ISignUpService
                 });
 
                 var token = await _signInManager.UserManager.FindByEmailAsync(newUser.Email);
-                if (token == null) return new Tuple<bool, string>(false, "");
+                if (token == null)
+                    return new ServiceResponse<string>(new Error(ErrorCodes.SignUpError, ""));
```
gonzalorom, line 63: Token could be null if and error happen validating the account. Is more an error in AWS Cognito server.

FiveRockingFingers/FRF.Core/Services/SignUpService.cs
```diff
@@ -58,13 +59,14 @@ public class SignUpService : ISignUpService
...
 
                 var result =
                     await _signInManager.PasswordSignInAsync(token, newUser.Password, false, false);
                 return result.Succeeded
-                    ? new Tuple<bool, string>(true, token.UserID)
-                    : new Tuple<bool, string>(false, "");
+                    ? new ServiceResponse<string>(token.UserID)
+                    : new ServiceResponse<string>(new Error(ErrorCodes.SignUpError, ""));
```
gonzalorom, line 69: Same as above, something happened with AWS Cognito server pool, and you couldn't accomplish the login.

FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
gonzalorom: Here you can have something like "_EmailAlreadyExist_" and "_AWSCognitoServerError_".

martinmena: This can mean multiple things. I'm not quite sure what to call it, and I'm not sure about how specific it should be. As far as I know, letting someone know that an email already exists in the database can be considered a security flaw, since it lets people know there's an account registered with that email, and then paired with brute force could let someone into the system. That's the reasoning behind this name, but this are all internal errors, business ones I think we called them, so I don't know how specific they should be.

matiasbeckerle: Good chat, I like it. Keep them coming.

Indeed exposing something like "email already exist" needs to be prevented if possible.

"Invalid credentials" is generic enough without telling if the username (or email) or password was wrong.

Regarding AWS Cognito error is telling about internal implementation details -- which we should avoid as well. Do we know why AWS Cognito can fail? Is this the error when the credentials are wrong? Or this means the AWS Cognito server is down, or something else?

martinmena: I think it would fail when AWS Cognito servers are down, but I'm not sure. Maybe @gonzalorom knows more.

gonzalorom: Like Matias said "Invalid credentials" is generic enough, because even if we have a password rule in the model binding this can cause a not succeeded response. With the AWS Cognito error, most of the time when the server is down, or you have an invalid secret_pool_id or just right after the creation of the user is done, you change the scope of your IAM AWS user, you will get this error. All these scenarios, in my opinion, are a bit rare and the end user can't do anything about it.

matiasbeckerle: Ok so as far as I understand then we need two errors: Invalid credentials and one for the AWS cognito error which should never happen. Maybe we can call this one AuthenticationServerCurrentlyUnavailable or something along those lines? And (I don't like it) but we should probably return a 500 error to consumers in that case.

401 for the invalid credentials, of course.

FiveRockingFingers/FRF.Core/Services/SignUpService.cs
```diff
@@ -26,7 +27,7 @@ public class SignUpService : ISignUpService
             _pool = pool;
         }
 
-        public async Task<Tuple<bool, string>> SignUpAsync(User newUser)
+        public async Task<ServiceResponse<string>> SignUpAsync(User newUser)
         {
             try
             {
```
gperelstein, line 33: Do we need this try/catch? We are not doing anything to if an exception occurs

martinmena: This is what's being done in the catch block:
`//throw exception from cognito user pool`
`throw new Exception("Sign up failed: " + e.Message);`

That's previous code and I wouldn't be able to tell you if it's needed or not at this point, maybe we're already checking for every chance of a cognito exception, maybe not. Probably we should ask @gonzalorom for this.
If we don't need this, I can remove it no problem.

But also, if we keep it, maybe it would be better to have a 
`return new ServiceResponse<string>(new Error(ErrorCodes.AuthenticationServerCurrentlyUnavailible, ""));`
instead of the _throw new Exception_

gperelstein: Returning a ServiceResponse object with an error if we don't know if we need or not this try-catch it's a great idea!

FiveRockingFingers/FRF.Core/Services/SignUpService.cs
```diff
@@ -38,7 +39,7 @@ public class SignUpService : ISignUpService
                 var response = await _userManager.CreateAsync(newCognitoUser, newUser.Password);
 
                 if (!response.Succeeded)
-                    return new Tuple<bool, string>(false, "");
+                    return new ServiceResponse<string>(new Error(ErrorCodes.InvalidCredentials, ""));
```
gperelstein, line 42: Everything looks great now! As a final comment, maybe it would be great to send a message when an error occurs. I don't know if it's redundant to send `new Error(ErrorCodes.InvalidCredentials, "Invalid Credentials")` but I'm thinking that we could use that message in the future in the controller to send something like BadRequest(response.Error.Message) when an error happens.

gperelstein (approved):
Everything looks great now! Just as a final comment, maybe it would be ok to send a message when an error occurs but I don't know if it's going

FiveRockingFingers/FRF.Core/Services/SignUpService.cs
martinmena: I thought about it too, and almost put the same "Invalid Credentials" message, but I don't know if it's user friendly enough for what you suggested. I'll think about it to see if I can come up with a different alternative, if that's ok.

gperelstein: You are right. Maybe something like **"There was an error with your E-Mail/Password combination"** or **"the Email or password is incorrect"** or **"You have entered an invalid username or password"**?

FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
martinmena: Done!

FiveRockingFingers/FRF.Core/Services/SignUpService.cs
martinmena: Done!

matiasbeckerle (approved):

PR #84, opened 2021-01-21
[Five-298] Agregar la regla no-console a eslint

gperelstein:
Cambios realizados

-Se agregó la regla no-console a eslint

### DIFF
--- a/FiveRockingFingers/FRF.Web/ClientApp/.eslintrc.json
+++ b/FiveRockingFingers/FRF.Web/ClientApp/.eslintrc.json
@@ -3,5 +3,8 @@
     "parserOptions": {
         "ecmaVersion": 6,
         "sourceType": "module"
+    },
+    "rules": {
+        "no-console": "warn"
     }
 }


gonzalorom (approved):
Very useful.

FiveRockingFingers/FRF.Web/ClientApp/.eslintrc.json
```diff
@@ -3,5 +3,8 @@
     "parserOptions": {
         "ecmaVersion": 6,
         "sourceType": "module"
+    },
+    "rules": {
+        "no-console": "warn"
```
matiasbeckerle, line 8: Error maybe? Will this prevent PRs from merging if they include a console.log?

gperelstein: We agree to add this rule as a warn because if we add it as an error, this will prevent running the app if someone adds a console.log() and we don't know if this function is going to be useful in the future something

matiasbeckerle: Ok makes sense

matiasbeckerle (approved):

martinmena (approved):

PR #87, opened 2021-01-25
[FIVE-295] Crear interfaz de precios de artefactos

gperelstein:
### Tareas realizadas

- Se creó la interfaz IArtifactPricing para el calcular el precio de un artifacto
- Se implementó la interfaz IArtifactPricing en el modelo de Core, Artifact
- Se creó la interfaz IBudgetService, el BudgetController, BudgetService y sus tests

### DIFF
--- a/FiveRockingFingers/FRF.Core/Models/IArtifactPricing.cs
+++ b/FiveRockingFingers/FRF.Core/Models/IArtifactPricing.cs
@@ -1,0 +1,7 @@
+﻿namespace FRF.Core.Models
+{
+    public interface IArtifactPricing
+    {
+        decimal GetPrice();
+    }
+}
--- a/FiveRockingFingers/FRF.Web/Controllers/BudgetController.cs
+++ b/FiveRockingFingers/FRF.Web/Controllers/BudgetController.cs
@@ -6,7 +6,7 @@ using System.Threading.Tasks;
 namespace FRF.Web.Controllers
 {
     [ApiController]
-    [Route("api/[controller]/[action]")]
+    [Route("api")]
     [Authorize]
     public class BudgetController : ControllerBase
     {
@@ -17,7 +17,7 @@ namespace FRF.Web.Controllers
             _budgetService = budgetService;
         }
 
-        [HttpGet("{projectId}")]
+        [HttpGet("/project/{projectId}/budget")]
         public async Task<IActionResult> GetBudget(int projectId)
         {
             var response = await _budgetService.GetBudget(projectId);
--- a/FiveRockingFingers/FRF.Web/Controllers/BudgetController.cs
+++ b/FiveRockingFingers/FRF.Web/Controllers/BudgetController.cs
@@ -1,0 +1,30 @@
+﻿using FRF.Core.Services;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using System.Threading.Tasks;
+
+namespace FRF.Web.Controllers
+{
+    [ApiController]
+    [Route("api/[controller]/[action]")]
+    [Authorize]
+    public class BudgetController : ControllerBase
+    {
+        private readonly IBudgetService _budgetService;
+
+        public BudgetController(IBudgetService budgetService)
+        {
+            _budgetService = budgetService;
+        }
+
+        [HttpGet("{projectId}")]
+        public async Task<IActionResult> GetBudget(int projectId)
+        {
+            var response = await _budgetService.GetBudget(projectId);
+
+            if (!response.Success) return NotFound();
+
+            return Ok(response.Value);
+        }
+    }
+}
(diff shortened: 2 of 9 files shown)


FiveRockingFingers/FRF.Web/Controllers/BudgetController.cs
```diff
@@ -0,0 +1,30 @@
+﻿using FRF.Core.Services;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using System.Threading.Tasks;
+
+namespace FRF.Web.Controllers
+{
+    [ApiController]
+    [Route("api/[controller]/[action]")]
```
matiasbeckerle, line 9: Don't. Please try to control your restful api endpoints as we previously talked, otherwise you will continue having odd endpoints.

FiveRockingFingers/FRF.Core/Models/IArtifactPricing.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace FRF.Core.Models
+{
+    public interface IArtifactPricing
```
matiasbeckerle, line 3: Interfaces are handy but I don't see any benefit on adding this one. Is it needed?

test/FRF.Core.Tests/Services/BudgetServiceTests.cs
```diff
@@ -0,0 +1,103 @@
...
+            _artifactsService.Setup(mock => mock.GetAllByProjectId(It.IsAny<int>()))
+                .ReturnsAsync(new ServiceResponse<List<Artifact>>(artifactsList));
+
+            //Act
+            var response = await _classUnderTest.GetBudget(project.Id);
+
+            //Assert
+            Assert.IsType<ServiceResponse<decimal>>(response);
+            Assert.True(response.Success);
+            Assert.Equal(response.Value, artifact.GetPrice());
```
matiasbeckerle, line 81: Remember that the order is: expected, actualResult.

test/FRF.Core.Tests/Services/BudgetServiceTests.cs
```diff
@@ -0,0 +1,103 @@
...
+            _artifactsService.Setup(mock => mock.GetAllByProjectId(It.IsAny<int>()))
+                .ReturnsAsync(new ServiceResponse<List<Artifact>>(new Error(ErrorCodes.ProjectNotExists, "[Mock] Message")));
+
+            //Act
+            var response = await _classUnderTest.GetBudget(projectId);
+
+            //Assert
+            Assert.IsType<ServiceResponse<decimal>>(response);
+            Assert.False(response.Success);
+            Assert.Equal(response.Error.Code, ErrorCodes.ProjectNotExists);
```
matiasbeckerle, line 99: Same as above. And same in the other testing file, please review the order on all of them.

FiveRockingFingers/FRF.Web/Controllers/BudgetController.cs
gperelstein: You are right! We already fix this, I hope is better now

test/FRF.Core.Tests/Services/BudgetServiceTests.cs
gperelstein: Done!

test/FRF.Core.Tests/Services/BudgetServiceTests.cs
gperelstein: Done! And we fix the tests in BudgetServiceTests

test/FRF.Core.Tests/Services/BudgetServiceTests.cs
gonzalorom: @matiasbeckerle Now we have created a new ticket FIVE-300 to go through all the tests and make the modifications.

FiveRockingFingers/FRF.Core/Models/IArtifactPricing.cs
egutbrod: The idea behind the interface, is to define the method contracts for artifact pricing, without thinking on the provider, the Abstract class here would be for AWS Artifacts, we can, on the future have other hierarchy for Google Cloud artifacts, or others, that also will need to implement these pricing methods, but not the same base attributes.

matiasbeckerle: Makes sense. Should `Artifact` be renamed to `AwsArtifact`?

matiasbeckerle (approved):

gonzalorom (approved):

FiveRockingFingers/FRF.Core/Models/IArtifactPricing.cs
gonzalorom: In my opinion we should leave it like this. The Next step is to made Artifact an Abstract class, so all the new artifacts like `AwsS3Artifact` or `GoogleContainerRegistryArtifact`, they will have to inherit from Artifact which is the parent class.

matiasbeckerle: Makes sense!

PR #92, opened 2021-01-26
[FIVE-299] (AwsArtifactsProvider) Revisar las URLs de endpoints para que sigan las buenas practicas de una api REST

gperelstein:
Cambios realizados

-Se modificó las URLs de los endpoints del controller AwsArtifactsProviderController para que sigan las buenas prácticas de una api REST

### DIFF
--- a/FiveRockingFingers/FRF.Web/ClientApp/src/services/AwsArtifactService.ts
+++ b/FiveRockingFingers/FRF.Web/ClientApp/src/services/AwsArtifactService.ts
@@ -2,7 +2,7 @@ import axios from 'axios'
 import { BASE_URL } from '../Constants';
 import KeyValueStringPair from '../interfaces/KeyValueStringPair';
 
-const AWS_ARTIFACTS_PROVIDER_URL = `${BASE_URL}awsArtifactsProvider`;
+const AWS_ARTIFACTS_PROVIDER_URL = `${BASE_URL}aws-artifacts-provider`;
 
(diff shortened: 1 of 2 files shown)


FiveRockingFingers/FRF.Web/ClientApp/src/services/AwsArtifactService.ts
```diff
@@ -2,25 +2,25 @@ import axios from 'axios'
 import { BASE_URL } from '../Constants';
 import KeyValueStringPair from '../interfaces/KeyValueStringPair';
 
-const AWS_ARTIFACTS_PROVIDER_URL = `${BASE_URL}AwsArtifactsProvider/`;
+const AWS_ARTIFACTS_PROVIDER_URL = `${BASE_URL}awsArtifactsProvider`;
```
matiasbeckerle, line 5: Maybe just `/aws`? Otherwise I would recommend the other style: `aws-artifacts-provider`

gperelstein: Maybe `/aws` is very general. I don't know if a URL like `api/aws/names` it's very self-explanatory of what represents. For now a leave it as `aws-artifacts-provider` but I don't think that is a good name either. Maybe something like `aws-services`? If I read the URL `api/aws-services/names`, I know exactly what I'm going to get

matiasbeckerle (approved):

PR #93, opened 2021-01-28
[FIVE-296] Crear implementación de precios para Artifact Custom

martinmena:
## Cambios realizados
- Se creó el modelo CustomArtifact y su mappeo con el modelo Artifact.
- El método GetPrice de Artifact pasó a ser virtual (y en CustomArtifact se lo sobreescribe).
- Se crearon dos métodos en ArtifactsService, mapArtifact que mappea un artefacto a su modelo correspondiente (por ahora solo CustomArtifact o Artifact), y mapArtifacts que cicla una lista de artefactos por el método mapArtifact. En base a eso, todos los métodos devuelven una lista de artefactos (o artefacto) igual que antes, pero con cada uno de ellos mappeado a su modelo específico.
- Se realizaron los cambios en el frontend para forzar la definición de un precio a la hora de crear un artefacto custom.


FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -32,6 +32,29 @@ private async Task<bool> DoArtifactsExist (IList<ArtifactsRelation> artifactsRel
             return artifactsRelationIds.Except(dbArtifactsId).Any();
         }
 
+        private Artifact mapArtifact(EntityModels.Artifact artifact)
```
matiasbeckerle, line 35: Odd function name casing. Same applies to the method below.

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -32,6 +32,29 @@ private async Task<bool> DoArtifactsExist (IList<ArtifactsRelation> artifactsRel
...
+        {
+            switch (artifact.Provider)
+            {
+                case "Custom":
```
matiasbeckerle, line 39: This string could be a constant or maybe provider be an Enum.

FiveRockingFingers/FRF.Core/Models/CustomArtifact.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace FRF.Core.Models
+{
+    public class CustomArtifact : Artifact
+    {
+        public override decimal GetPrice()
```
matiasbeckerle, line 5: Some unit tests for this method will be handy.

matiasbeckerle (approved):

FiveRockingFingers/FRF.Core/Models/CustomArtifact.cs
martinmena: Does this mean that we should have a Models folder inside the FRF.Core.Tests with a test for every type of artifact we create? In this case it would be _FRF.Core.Tests -> Models -> CustomArtifactTests.cs_ and it would only test this one method, right?
And when I say every type of artifact I mean that we are going to have a specific model for every type of service that the provider offers (in the case of AWS we're going to have a S3 model, EC2 model, API Gateway model, etc) with the specific method to calculate the price according to the service, and every one of those should also have it's own tests, right?
_FRF.Core.Tests -> Models -> AWS -> S3Tests.cs_
_FRF.Core.Tests -> Models -> AWS -> EC2Tests.cs_
_FRF.Core.Tests -> Models -> AWS -> ApiGatewayTests.cs_

gperelstein (approved):

FiveRockingFingers/FRF.Core/Models/CustomArtifact.cs
matiasbeckerle: Yes, exactly as you described.

But to be honest I'm not sure what "custom" means in your case, sounds like a fake class while you prepare to have the real ones. So if this class is just a placeholder for what's next, then feel free to ignore. Once you have the real ones, don't forget to unit test these type of logic though.

martinmena: Nope, custom is an actual type of artifact, that isn't a part of any provider, but could represent something involved in the project and have relations with the other artifacts. Maybe it's a little hard to explain, but the important part is that it is a real class, so I'll create the tests.

matiasbeckerle: Awesome, thanks!

gonzalorom (approved):
Just the naming in `mapArtifacts` and `mapArtifact`.

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
martinmena: Sounds much better, but maybe it would be best to find another solution in that case. We already have a constant in the frontend with the definition of _Providers_ (as far as I know, it's the only definition there is)
_export const PROVIDERS = ['AWS', 'Custom'];_
From that definition we get the options for the form when we create an artifact, and then the string ('AWS' or 'Custom') is saved on the database.
If we create another definition in the backend, and tomorrow we add another provider we're going to have to update both constants, or will probably get some errors working with that new provider.
What I'm trying to say is that it may be better to have only one definition for providers, and both backend and frontend work with that, but I don't know where that could be done. (I was thinking that we could create another table in the database with the providers (Id - Name - Something else if needed) and make a new service/controller to work with it, but it might be too much for something this simple)

matiasbeckerle: Hehe I understand your point. Yes, you could create something on the backend (like a providers table) that could fill any consumer and have only 1 source of truth.

On the other hand, maybe that's not worthy and that's fine too. Having to declare constants on both ends is acceptable. Remember that in the case you need to add 1 provider, all applications probably requires modifications, besides adding a constant value.

martinmena: Done! I don't know if it's the best way to do it, so let me know if you suggest any more changes

FiveRockingFingers/FRF.Core/Constants.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace FRF.Core
+{
+    public static class Constants
```
matiasbeckerle, line 3: I believe the filename and this class should be `ArtifactTypes` instead so in code you can do something like... `ArtifactTypes.Aws`.

FiveRockingFingers/FRF.Core/Constants.cs
```diff
@@ -0,0 +1,11 @@
...
+    {
+        public static class ArtifactTypes
+        {
+            public const string CUSTOM = "Custom";
+            public const string AWS = "AWS";
```
matiasbeckerle, line 8: I'm OK whatever style you prefer but I want to point out that MSFT constants are usually `PascalCase` instead. In your case, Custom and Aws.

matiasbeckerle (approved):

FiveRockingFingers/FRF.Core/Constants.cs
martinmena: Awesome! Thanks for the feedback!

FiveRockingFingers/FRF.Core/Constants.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace FRF.Core
+{
+    public static class Constants
+    {
+        public static class ArtifactTypes
+        {
+            public const string CUSTOM = "Custom";
```
gonzalorom: It will be handy for me because I need to implement some constants as well 💪.

gonzalorom (approved):

FiveRockingFingers/FRF.Core/Constants.cs
gperelstein: Yes, this is going to be very useful!

PR #95, opened 2021-02-01
[Five-301](XmlValidation) Modificar como se guardan los precios dentro de las settings de un artefacto

gperelstein:
Cambios realizados

-Se creó el método ValidateSettings en el servicio ArtifactsService para validar la propiedad settings de un artefacto según un esquema de validación
-Se creó el esquema de validación CustomArtifact.xsd para la validación de la propiedad settings de artefactos Custom
-Se modificó el archivo FRF.Web.csproj para incluir el archivo de validación CustomArtifact.xsd en el proyecto

### DIFF
--- a/FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
+++ b/FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
@@ -16,7 +16,8 @@
         // User errors
         public const int InvalidCredentials = 20;
         public const int AuthenticationServerCurrentlyUnavailable = 21;
-        
+        public const int UserNotExists = 22;
+
         // External errors
         public const int AmazonApiError = 30;
     }
--- a/FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
+++ b/FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
@@ -12,5 +12,9 @@
         public const int RelationAlreadyExisted = 10;
         public const int RelationNotValid = 11;
         public const int RelationNotExists = 12;
+
+        // User errors
+        public const int InvalidCredentials = 20;
+        public const int AuthenticationServerCurrentlyUnavailable = 21;
     }
 }
--- a/FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
+++ b/FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
@@ -22,6 +22,6 @@
         public const int AmazonApiError = 30;
 
         //Artifacts errors
-        public const int InvalidSettings = 40;
+        public const int InvalidArtifactSettings = 40;
     }
 }
--- a/FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
+++ b/FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
@@ -20,5 +20,8 @@
 
         // External errors
         public const int AmazonApiError = 30;
+
+        //Artifacts errors
+        public const int InvalidSettings = 40;
     }
 }
--- a/FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
+++ b/FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
@@ -16,5 +16,8 @@
         // User errors
         public const int InvalidCredentials = 20;
         public const int AuthenticationServerCurrentlyUnavailable = 21;
+        
+        // External errors
+        public const int AmazonApiError = 30;
     }
 }
--- a/FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
+++ b/FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
@@ -11,5 +11,6 @@
         // Relation errors
         public const int RelationAlreadyExisted = 10;
         public const int RelationNotValid = 11;
+        public const int RelationNotExists = 12;
     }
 }
--- a/FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
+++ b/FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
@@ -20,11 +20,13 @@ namespace FRF.Core.Services
     {
         private readonly DataAccessContext _dataContext;
         private readonly IMapper _mapper;
+        private readonly ISettingsValidator _settingsValidator;
 
-        public ArtifactsService(DataAccessContext dataContext, IMapper mapper)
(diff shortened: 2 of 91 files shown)


FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -128,6 +132,11 @@ public async Task<ServiceResponse<Artifact>> Save(Artifact artifact)
                 return new ServiceResponse<Artifact>(new Error(ErrorCodes.ArtifactTypeNotExists, $"There is no artifact type with Id = {artifact.ArtifactTypeId}"));
             }
 
+            if(artifact.Provider == "Custom" && !ValidateSettings(artifact.Settings))
```
gperelstein, line 135: Here I have to implement the constants that @martinmena created. I'm waiting for the merge of that PR

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -128,6 +132,11 @@ public async Task<ServiceResponse<Artifact>> Save(Artifact artifact)
...
+            {
+                return new ServiceResponse<Artifact>(new Error(ErrorCodes.ArtifactTypeNotExists, $"Settings are incorrect"));
```
matiasbeckerle, line 137: _ArtifactTypeNotExists_ are you sure that's the case here?

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -311,5 +320,36 @@ public async Task<ServiceResponse<ArtifactsRelation>> DeleteRelationAsync(Guid a
 
             return new ServiceResponse<IList<ArtifactsRelation>>(artifactsRelationsNew);
         }
+
+        private bool ValidateSettings(XElement settings)
+        {
+            var isAnError = true;
+
+            string rootPath = Path.GetDirectoryName(System.Reflection.Assembly.GetExecutingAssembly().Location);
```
matiasbeckerle, line 328: Is reflection needed? Can't you use something else, less expensive?

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gperelstein: The constants have been already implemented

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -335,5 +344,36 @@ public async Task<ServiceResponse<ArtifactsRelation>> DeleteRelationAsync(Guid a
 
             return new ServiceResponse<IList<ArtifactsRelation>>(artifactsRelationsNew);
         }
+
+        private bool ValidateSettings(XElement settings)
```
matiasbeckerle, line 348: Similar to what I mentioned in another PR this code could live in another place, some sort of validation class or helper.

FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
```diff
@@ -20,5 +20,8 @@ public static class ErrorCodes
 
         // External errors
         public const int AmazonApiError = 30;
+
+        //Artifacts errors
+        public const int InvalidSettings = 40;
```
matiasbeckerle, line 25: _InvalidSettings_ is quite generic so it gets some sort of prefix or you don't make it an artifact exclusive. You will have to pick 1 path.

matiasbeckerle (approved):

FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
gperelstein: When I say settings I refer to the artifact's property called settings

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gperelstein: I move it to a new Class called `SettingsValidator`

FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
matiasbeckerle: This is my reasoning behind:

What will you do next time you have to report an error related to an invalid setting? For another entity, or maybe even for a completely different use case. Will you use this constant? Or maybe this constant name should be more specific?

FiveRockingFingers/FRF.Core/XmlValidation/SettingsValidator.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Text;
+using System.Xml;
+using System.Xml.Linq;
+using System.Xml.Schema;
+
+namespace FRF.Core.XmlValidation
+{
+    public class SettingsValidator
+    {
+        public static bool ValidateSettings(XElement settings)
```
matiasbeckerle, line 13: Just a note: by making this static, this method will be unmockable. For example, when you try to unit test the service method that calls this function.

That doesn't necessary mean it's something bad. It just mean that you cannot control the output of this function when unit testing. Keep that in mind while working on tests. If you reach some sort of scenario where you don't have control, then maybe this method should not be static?

This comment doesn't mean you have to change this code, is just an observation.

matiasbeckerle (approved):

FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
gperelstein: I understand and it makes a lot of sense. I think that leaving a generic error name and use it for different use cases is not a good idea because if do that, every time this error appears we are not going to know what the problem was exactly. I think that a name like `InvalidArtfifactSettings` expresses very well what was the problem, "The artifact's settings that you are sending are invalid". Or even I could combine my two ideas in one in something like `InvalidArtfifactSettingsFormat` but maybe this error name is too long.

matiasbeckerle: Good reasoning. That's what I wanted to achieve 😁

gonzalorom (approved):

martinmena (approved):

FiveRockingFingers/FRF.Core/XmlValidation/SettingsValidator.cs
gperelstein: You are right. I chose to make this method not static and create tests specifically for this method

PR #96, opened 2021-02-01
[FIVE-297] (Models) Crear implementacion de precios para Artifact AWS S3

gonzalorom:
### Tareas realizadas:
- Se creó clase AwsS3Artifact.
- Se modifica método GetPrice en Models/Artifact a tipo virtual.
- Se creó sobrescribe método GetPrice en AwsS3Artifact el cual implementa GetIntelligentPrice y GetStandardPrice.
- Se creó método GetS3ProductsAsync y GetS3IntelligentTieringDetailList.

### DIFF
--- a/FiveRockingFingers/FRF.Core/ArtifactTypes.cs
+++ b/FiveRockingFingers/FRF.Core/ArtifactTypes.cs
@@ -5,4 +5,25 @@
         public const string Custom = "Custom";
         public const string Aws = "AWS";
     }
+
+    public static class AwsS3Constants
+    {
+        public const string Service = "AmazonS3";
+        public const string WriteFrequentGroup = "S3-API-Tier1";
+        public const string RetrieveFrequentGroup = "S3-API-Tier2";
+
+        public const string StandardInfrequentAccessProduct = "Standard - Infrequent Access";
+        public const string IntelligentInfrequentAccessProduct = "Intelligent-Tiering Infrequent Access";
+        public const string IntelligentFrequentAccessProduct = "Intelligent-Tiering Frequent Access";
+        public const string WriteInfrequentGroup = "S3-API-SIA-Tier1";
+        public const string RetrieveInfrequentGroup = "S3-API-SIA-Tier2";
+
+        public const string InfrequentAccessProduct = "One Zone - Infrequent Access";
+        public const string WriteOneZoneInfrequentGroup = "S3-API-ZIA-Tier1";
+        public const string RetrieveOneZoneInfrequentGroup = "S3-API-ZIA-Tier2";
+
+        public const string AutomationObjectCountFee = "S3-Monitoring and Automation-ObjectCount";
+        public const string IntelligentTieringProduct = "Intelligent-Tiering";
+
+    }
 }
--- a/FiveRockingFingers/FRF.Core/ArtifactTypes.cs
+++ b/FiveRockingFingers/FRF.Core/ArtifactTypes.cs
@@ -1,0 +1,8 @@
+﻿namespace FRF.Core
+{
+    public static class ArtifactTypes
+    {
+        public const string Custom = "Custom";
+        public const string Aws = "AWS";
+    }
+}
--- a/FiveRockingFingers/FRF.Core/ArtifactTypes.cs
+++ b/FiveRockingFingers/FRF.Core/ArtifactTypes.cs
@@ -5,25 +5,4 @@
         public const string Custom = "Custom";
         public const string Aws = "AWS";
     }
-
-    public static class AwsS3Constants
-    {
-        public const string Service = "AmazonS3";
-        public const string WriteFrequentGroup = "S3-API-Tier1";
-        public const string RetrieveFrequentGroup = "S3-API-Tier2";
-
-        public const string StandardInfrequentAccessProduct = "Standard - Infrequent Access";
-        public const string IntelligentInfrequentAccessProduct = "Intelligent-Tiering Infrequent Access";
-        public const string IntelligentFrequentAccessProduct = "Intelligent-Tiering Frequent Access";
-        public const string WriteInfrequentGroup = "S3-API-SIA-Tier1";
-        public const string RetrieveInfrequentGroup = "S3-API-SIA-Tier2";
-
-        public const string InfrequentAccessProduct = "One Zone - Infrequent Access";
-        public const string WriteOneZoneInfrequentGroup = "S3-API-ZIA-Tier1";
-        public const string RetrieveOneZoneInfrequentGroup = "S3-API-ZIA-Tier2";
-
-        public const string AutomationObjectCountFee = "S3-Monitoring and Automation-ObjectCount";
-        public const string IntelligentTieringProduct = "Intelligent-Tiering";
-
-    }
 }
--- a/FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
+++ b/FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
@@ -26,7 +26,7 @@ namespace FRF.Core
                 .ForMember(dest => dest.Id, act => act.Ignore())
                 .ForMember(dest => dest.ArtifactType, act => act.Ignore())
                 .ForMember(dest => dest.Project, act => act.Ignore());
-            CreateMap<DataAccess.EntityModels.Artifact, AwsS3Artifact>()
+            CreateMap<DataAccess.EntityModels.Artifact, AwsS3>()
                 .ForMember(dest => dest.Id, act => act.Ignore())
                 .ForMember(dest => dest.ArtifactType, act => act.Ignore())
                 .ForMember(dest => dest.Project, act => act.Ignore())
@@ -53,7 +53,7 @@ namespace FRF.Core
                         decimal.Parse(
                             ar.Settings.Element("product3").Element("pricingDimensions").Element("pricePerUnit").Value,
                             NumberStyles.Float)));
-            CreateMap<AwsS3Artifact, DataAccess.EntityModels.Artifact > ()
+            CreateMap<AwsS3, DataAccess.EntityModels.Artifact > ()
                 .ForMember(dest => dest.Id, act => act.Ignore())
                 .ForMember(dest => dest.ArtifactType, act => act.Ignore())
                 .ForMember(dest => dest.Project, act => act.Ignore());
--- a/FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
+++ b/FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
@@ -1,5 +1,8 @@
-﻿using AutoMapper;
+﻿using System;
+using System.Globalization;
+using AutoMapper;
 using FRF.Core.Models;
+using FRF.Core.Models.AwsArtifacts;
 
 namespace FRF.Core
 {
@@ -20,6 +23,37 @@ namespace FRF.Core
                 .ReverseMap();
             CreateMap<DataAccess.EntityModels.Artifact, Models.Artifact>()
                 .ReverseMap()
+                .ForMember(dest => dest.Id, act => act.Ignore())
+                .ForMember(dest => dest.ArtifactType, act => act.Ignore())
+                .ForMember(dest => dest.Project, act => act.Ignore());
+            CreateMap<DataAccess.EntityModels.Artifact, AwsS3Artifact>()
+                .ForMember(dest => dest.Id, act => act.Ignore())
+                .ForMember(dest => dest.ArtifactType, act => act.Ignore())
+                .ForMember(dest => dest.Project, act => act.Ignore())
+                .ForMember(dest => dest.WriteRequestsUsed, act =>
+                    act.MapFrom(ar => ar.Settings.Element("writeRequestsUsed").Value))
+                .ForMember(dest => dest.RetrieveRequestsUsed, act =>
+                    act.MapFrom(ar => ar.Settings.Element("retrieveRequestsUsed").Value))
+                .ForMember(dest => dest.StorageUsed, act =>
+                    act.MapFrom(ar => ar.Settings.Element("storageUsed").Value))
+                .ForMember(dest => dest.InfrequentAccessMultiplier, act =>
+                    act.MapFrom(ar => ar.Settings.Element("infrequentAccessMultiplier").Value))
+                .ForMember(dest => dest.StoragePrice, act =>
+                    act.MapFrom(ar =>
+                        Convert.ToDecimal(
+                            ar.Settings.Element("product1").Element("pricingDimensions").Element("range0")
+                                .Element("pricePerUnit").Value, CultureInfo.InvariantCulture)))
+                .ForMember(dest => dest.WriteRequestsPrice, act =>
+                    act.MapFrom(ar =>
+                        decimal.Parse(
+                            ar.Settings.Element("product2").Element("pricingDimensions").Element("pricePerUnit").Value,
+                            NumberStyles.Float)))
+                .ForMember(dest => dest.RetrieveRequestsPrice, act =>
+                    act.MapFrom(ar =>
+                        decimal.Parse(
+                            ar.Settings.Element("product3").Element("pricingDimensions").Element("pricePerUnit").Value,
+                            NumberStyles.Float)));
+            CreateMap<AwsS3Artifact, DataAccess.EntityModels.Artifact > ()
                 .ForMember(dest => dest.Id, act => act.Ignore())
                 .ForMember(dest => dest.ArtifactType, act => act.Ignore())
                 .ForMember(dest => dest.Project, act => act.Ignore());
--- a/FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
+++ b/FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
@@ -24,11 +24,9 @@ namespace FRF.Core
             CreateMap<DataAccess.EntityModels.Artifact, Models.Artifact>()
                 .ReverseMap()
                 .ForMember(dest => dest.Id, act => act.Ignore())
-                .ForMember(dest => dest.ArtifactType, act => act.Ignore())
                 .ForMember(dest => dest.Project, act => act.Ignore());
             CreateMap<DataAccess.EntityModels.Artifact, AwsS3>()
                 .ForMember(dest => dest.Id, act => act.Ignore())
-                .ForMember(dest => dest.ArtifactType, act => act.Ignore())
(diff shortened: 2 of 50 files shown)


FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsS3Artifact.cs
```diff
@@ -0,0 +1,75 @@
...
+                        .Value,
+                    out var beginRangeLast)) return 0;
+
+            decimal storageFrequentAccessCost;
+            const decimal InfrequentAccessPrice = 0.0125m;
+
+            var storageInfrequentAccessMultiplier = (int) InfrequentAccessMultiplier / 100;
+            var storageFrequentAccessMultiplier = 1 - storageInfrequentAccessMultiplier;
+            var storageFrequentAccessUsed = StorageUsed * storageFrequentAccessMultiplier;
+            if (storageFrequentAccessUsed <= endRange1)
```
gperelstein, line 50: I have a question. We agreed that if the endRange is "Inf" we are going to set it as -1. That is taken into consideration here? Of course that we can change how endRange is set in the case of "Inf"

gperelstein (approved):
Great work!

FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
```diff
@@ -23,6 +26,37 @@ public AutoMapperProfile()
                 .ForMember(dest => dest.Id, act => act.Ignore())
                 .ForMember(dest => dest.ArtifactType, act => act.Ignore())
                 .ForMember(dest => dest.Project, act => act.Ignore());
+            CreateMap<DataAccess.EntityModels.Artifact, AwsS3Artifact>()
+                .ForMember(dest => dest.Id, act => act.Ignore())
```
matiasbeckerle, line 30: Why do you ignore these properties?

FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsS3Artifact.cs
```diff
@@ -0,0 +1,75 @@
...
+            if (!int.TryParse(
+                    Settings.Element("product1").Element("pricingDimensions").Element("range2").Element("endRange")
+                        .Value,
+                    out var endRange1) ||
+                !int.TryParse(
+                    Settings.Element("product1").Element("pricingDimensions").Element("range1").Element("endRange")
+                        .Value,
+                    out var endRange2) ||
+                !int.TryParse(
+                    Settings.Element("product1").Element("pricingDimensions").Element("range0").Element("beginRange")
```
matiasbeckerle, line 40: Can't _pricingDimensions_ be part of the class as you did with the others?

FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsS3Artifact.cs
gonzalorom: In all the test that I made `range0` had and Inf (-1) for the endRange for that reason, I only check for the last beginRange in line 56. That's because just the storage price (product 1) can have three values arranged as follows: 
range 0 last: 500 to Infinite
range 1 middle:  450 to 500
range 2 begin: 0 to 450

FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
gonzalorom: Here all the artefacts are already uniques and we only use the Id when interactions with the database are done. The same happens above between _**EntityModels.Artifact and Artifact**_.

matiasbeckerle: Hmm I'm not sure I follow. Isn't `AwsS3Artifact` going to represent an entity that is on the database?

FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsS3Artifact.cs
gonzalorom: That's a good one, one of the most popular topics when designing this class was, what properties should the class have. Since all these "settings" are in "settings" and they were going to be repeated. PriceDimension is an array of complex objects distinct from others that are primitive data types.
So to avoid long lines of code (... Element (). Element (). Value) I copy some of the data in the properties of the AwsS3Artifact object.
With this in mind, would I have to do the same with PriceDimension?

FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
gonzalorom: No, because all the necessary information is inside of `Artifact.Settings`.

matiasbeckerle: I still don't understand the idea, sorry. What's `DataAccess.EntityModels.Artifact` and what's `AwsS3Artifact`?

FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsS3Artifact.cs
matiasbeckerle: I'm not sure but to me this looks like a code smell.

Remember that when you don't have primitive data types you can create your own classes. Will that help you in this regard?

FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
gonzalorom: There is no reason to be sorry, it was also difficult for me to understand it at once. 

AwsS3Artifact heritage from Artifact but doesn't have properties, whats only do is implement `GetPrice()`. All the information is inside _Settings_ which is in the parent class "**Artifact**": **_in XML format_**. I can leave AwsS3Artifact just with the overwrite of `GetPrice()` and It will work as well, but to make the code easier to read I encapsulate some of the value from _Settings_ inside props like `RetrieveRequestsPrice` or `StorageUsed`.

At the end of the day when you want to obtain the price of an artefact from the database, you will verify the name and the provider that it has and consequently, of those parameters, you will be able to map an **AwsS3Artifact** or an **AzureBlobStorageArtifact** to implement the corresponding GetPrice logic. It is exactly the same as with ArtifactCustom. 

I hope I was a little clearer with my explanation.

FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsS3Artifact.cs
```diff
@@ -0,0 +1,75 @@
+﻿using System.Globalization;
+
+namespace FRF.Core.Models.AwsArtifacts
+{
+    public class AwsS3Artifact : Artifact
```
gperelstein, line 5: What do you think about simplifying the name of this class to _S3_? I don't know if is necessary to add _Aws_ or _Artifact_ because this class is already inside a folder called AwsArtifacts

FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
gperelstein: Exactly as @gonzalorom  explained. The same was done by @martinmena in #93

FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsS3Artifact.cs
gperelstein: Excellent!

FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsS3Artifact.cs
gonzalorom: I like the idea, looks clean **S3**, same will happen for the rest RDS, DynamoDB and so on. But at the same time, it feels a bit odd if you read it in the middle of the code. What do you think @martinmena?

FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
martinmena: I don't know if this is exactly what matias meant, but I wanted to add to the idea of "ignoring" this properties, because now I'm second guessing myself. When we first ignored the "Id" property, it was because when we save an artifact in the database we want to avoid any Id predefined by the user, and assign a new unique one when the artifact is created, that's why we have an ignore in the Model.Artifact -> EntityModels.Artifact if I'm correct. We don't use the same map in any other case (at least I think we don't).
I don't know if we even need the reverse map (the one below this one) and there's no need to ignore the Id now that the artifact already exists in the database.
And just to finish my comment, yes, I did this same thing for the custom artifact, and I think we should keep it consistent if they behave the same, so if we remove the reverse map, and the ignoring the id, we should change the other mapping also.

FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsS3Artifact.cs
martinmena: Could work. My only concern is, as you pointed out, that in can be a little odd to read, and I think specially if there are other providers with the same artifact name. Let's say Aws has an S3, and also Azure. In that case, if you use both of them in the same chunk of code, it might get a little confusing I guess, but I'm not sure.

FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
matiasbeckerle: Thanks for the explanations and yes, my question is related to what @martinmena mentions. I think I get the point of why you need these classes where each has it's own _GetPrice_ method. What I'm curious about is why you set to ignore all these other properties like ID and more. If what you think it's fine and make sense on your understanding of the business then fine. From the point of view of an outsider, ignoring the ID and other properties looks suspicious UNLESS you tell me "yeah that's because these classes doesn't represent an actual record on the database, are just to make some business stuff, don't worry". But you aren't telling me that either 😁

So... TL;DR: I still don't understand the whole scope of these classes and that's fine since I can't invest much time in the project. Just in case, I mentioned hey folks, you sure you want to ignore IDs and other props? Time will tell if you needed them or not.

Bonus track: always be 100% sure of what you are using automapper for. The precise moment where you not understand exactly what are you mapping, boom 💥 you have a bug. With automapper, it's that easy. And don't think .ReverseMap is the best thing in the world, be careful when you use it too.

FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsS3Artifact.cs
gonzalorom: Agreed. Let's move on whit **_AwsS3_** looks like a middle point. I know that we already are calling this artifact from **Aws**Artifacts. But in this way, it looks shorter and clearer in the context where it can be called.

FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsS3Artifact.cs
gonzalorom: Done. I made a POCO obj to handle all this. 
PS: _***Code smell*** This sentence opened the door for me to a world of very interesting blogs._

FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
gperelstein: _"yeah that's because these classes doesn't represent an actual record on the database, are just to make some business stuff, don't worry"_  I think that phrase is way more clear of what we are trying to say. As you say we only going to use these classes to calculate the price of an artifact but we are not going to store them in the database. The values that these classes need to calculate the price are going to be stored in the settings of the artifact. These classes only need to take those parameters and calculate the price. And as you say, every type of artifact is going to have its own method `GetPrice` to calculate the price.
For that, I don't think we need the .ReverseMap()

FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsS3Artifact.cs
gperelstein: It looks much better and as you say is very understandable!

FiveRockingFingers/FRF.Core/ArtifactTypes.cs
```diff
@@ -5,4 +5,25 @@ public static class ArtifactTypes
         public const string Custom = "Custom";
         public const string Aws = "AWS";
     }
+
+    public static class AwsS3Constants
```
matiasbeckerle, line 9: Filename is _ArtifactTypes_. I believe you intend to place these into a new file maybe?

FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
matiasbeckerle: Cool

FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsS3.cs
```diff
@@ -0,0 +1,69 @@
...
+            var doc = new XmlDocument();
+            if (Settings.Element("product1") == null) return 0;
+            
+            doc.LoadXml(Settings.Element("product1").Element("pricingDimensions").ToString());
+            var priceDimensionsJson = JsonConvert.SerializeXmlNode(doc);
+            PricingDimensions = JsonConvert.DeserializeObject<PricingDimensionPocos>(priceDimensionsJson);
+
+            return InfrequentAccessMultiplier != 0 && InfrequentAccessMultiplier != null ?
+                GetIntelligentPrice() : GetStandardPrice();
+        }
```
matiasbeckerle, line 28: You **have to** test the hell out of this method hehe.

matiasbeckerle (commented):
I think these changes needs tons of tests for the price method.

FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsS3.cs
gonzalorom: That's right, I'm working on it.

FiveRockingFingers/FRF.Core/ArtifactTypes.cs
gonzalorom: I will move it to a different file, Is okay to keep the name AwsS3Constants? Thank you.

matiasbeckerle: I'm not much of a fan of adding words that represents something of the code (that the values are constants, for example). Maybe `AwsS3Descriptions`?

gonzalorom: I remember that you told me once about it. What you suggest seems better to me.

FiveRockingFingers/FRF.Core/AwsS3Descriptions.cs
```diff
@@ -0,0 +1,22 @@
+﻿namespace FRF.Core
```
gperelstein, line 1: Maybe this file should go in a folder like AwsServicesDescription or something like that, what do you think?

gonzalorom: That's a good idea. We can have something like "Descriptions" or "AwsDescriptions" as a folder. So all this kind of class can go there.

matiasbeckerle: Could be. I would advice to have this code maturing so you will clearly see in the near future _where_ to put this class and probably others.

test/FRF.Core.Tests/Models/AwsArtifacts/AwsS3Test.cs
```diff
@@ -0,0 +1,315 @@
...
+    {
+        private readonly AwsS3 _classUnderTest;
+
+        public AwsS3Test()
+        {
+            _classUnderTest = new AwsS3();
+        }
+
+        [Fact]
+        public void GetPrice_WhenIsIntelligentTiering_ReturnDecimalPrice()
```
matiasbeckerle, line 18: In the future if you reach an issue / error / exception calculating a price I would advice creating a test first. That way you easily reproduce the issue. Then you fix the code and the test shall pass. I'm telling you this now because the code calculating the price looks fragile because all the casting/conversions/serialization happening.

matiasbeckerle (approved):

FiveRockingFingers/FRF.Core/AwsS3Descriptions.cs
gonzalorom: Good advice, I will let the class where it is now.

martinmena (approved):

test/FRF.Core.Tests/Models/AwsArtifacts/AwsS3Test.cs
gonzalorom: Yes, your Matías is right. As I was writing the tests and the code, I found myself refactoring sectors that I had already tested. In this case, what do you think of wrapping the price calculation in a try / catch and returning a response similar to ServiceError?

matiasbeckerle: I'm not exactly sure. At this stage I have the feeling you better find and fix bugs faster than hiding with exceptions. If you throw exceptions, you have a way to continue the flow of the request? Your call.

gonzalorom: You will not have the possibility to continue the flow of the request. You would have to resubmit a new one. I will leave it like this, so like you said, we can act fast against possibles bugs.

test/FRF.Core.Tests/Models/AwsArtifacts/AwsS3Test.cs
```diff
@@ -0,0 +1,315 @@
...
+            _classUnderTest.StorageUsed = storageUsed;
+            _classUnderTest.RetrieveRequestsUsed = retrieveRequestsUsed;
+            _classUnderTest.WriteRequestsUsed = writeRequestsUsed;
+            _classUnderTest.InfrequentAccessMultiplier = infrequentAccessMultiplier;
+            
+            // Act
+            var result = _classUnderTest.GetPrice();
+            // Assert
+            Assert.IsType<decimal>(result);
+            Assert.Equal(FinalCost,result);
```
gperelstein, line 47: I'm getting an error here. It's expecting 17.3249920 and is receiving 19.0473600

gonzalorom: Nice catch, I had changed the end ranges and I forget to fix the result. This is when I miss CI.

PR #98, opened 2021-02-03
[FIVE-303] (Backend) Modificar tablas Artifacts, ArtifactType, y Providers en Base de Datos

martinmena:
## Descripción
En base a la necesidad de guardar la información acerca de que artefacto específico se está almacenando para luego poder instanciar un objeto específico para cada uno con su cálculo del precio, se resolvió modificar la base de datos como se puede ver en la imagen (lo que se encuentra en rojo es eliminación y en verde lo nuevo). Con estas modificaciones, todo artefacto tiene asociado un ArtifactType (no se estaba usando ese campo hasta el momento), los ArtifactType tienen un nombre que representa el servicio específico, y tienen asociado un proveedor (de haberse mantenido esa columna en Artifacts no había una relación directa entre los proveedores y sus servicios o artefactos, por lo que la información podía volverse incoherente).
ChangesDB

## Tareas a realizar
- Crear la tabla **Providers** con un campo “_Id_” como clave primaria, y un campo “_Name_” para almacenar los nombres (varchar).
- En la tabla **Artifacts** se debe eliminar la columna “_Provider_”.
- En la tabla **ArtifactType** se deben agregar las columnas “_RequiredFields_” (nvarchar) y “_ProviderId_” (foreign key de la tabla Providers).

## Cambios realizados
Además de realizar los cambios en la base de datos como se específica más arriba, se realizaron los cambios necesarios para que la aplicación pudiera seguir funcionando de manera correcta. Esto significa actualizar los EntityModels, Models y Dtos de Artefactos y ArtifactTypes, además de un EntityModel, Model y Dto para trabajar con los Providers (junto con sus mapeos necesarios).

Por último, se actualizó el servicio de artefactos, y los test de dicho servicio, del servicio de presupuestos, y de el controller de artefactos. Durante las pruebas del servicio de artefactos se detectaron problemas con las respuestas del Update y Save y se corrigieron (esto se daba porque se intentaba instanciar un objeto específico en base al proveedor del artefacto, pero no se tenía esa información luego de crearlo o modificarlo).

En otra historia se encontrarán las modificaciones necesarias sobre el FrontEnd para adaptarlo a estos cambios.

### DIFF
--- a/FiveRockingFingers/FRF.DataAccess/EntityModels/Provider.cs
+++ b/FiveRockingFingers/FRF.DataAccess/EntityModels/Provider.cs
@@ -1,0 +1,11 @@
+﻿using System.ComponentModel.DataAnnotations;
+
+namespace FRF.DataAccess.EntityModels
+{
+    public class Provider
+    {
+        [Key]
+        public int Id { get; set; }
+        public string Name { get; set; }
+    }
+}
--- a/FiveRockingFingers/FRF.DataAccess/EntityModels/Provider.cs
+++ b/FiveRockingFingers/FRF.DataAccess/EntityModels/Provider.cs
@@ -1,4 +1,5 @@
-﻿using System.ComponentModel.DataAnnotations;
+﻿using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
 
 namespace FRF.DataAccess.EntityModels
 {
@@ -7,5 +8,6 @@ namespace FRF.DataAccess.EntityModels
         [Key]
         public int Id { get; set; }
         public string Name { get; set; }
+        public ICollection<ArtifactType> ArtifactType { get; set; }
     }
 }
--- a/FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
+++ b/FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
@@ -35,6 +35,7 @@ namespace FRF.Core
                 .ForMember(dest => dest.Id, act => act.Ignore());
             CreateMap<DataAccess.EntityModels.ArtifactsRelation, ArtifactsRelation>()
                 .ReverseMap();
+            CreateMap<DataAccess.EntityModels.Provider, Provider>();
         }
     }
 }
--- a/FiveRockingFingers/FRF.Core/Models/Artifact.cs
+++ b/FiveRockingFingers/FRF.Core/Models/Artifact.cs
@@ -7,7 +7,6 @@ namespace FRF.Core.Models
     {
         public int Id { get; set; }
         public string Name { get; set; }
-        public string Provider { get; set; }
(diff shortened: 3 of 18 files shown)


gperelstein (approved):
Excellent!

matiasbeckerle (approved):
Changes looks good but I would like to leave a comment. I bet you have been thinking hard on this and I believe this is an important decision with no return (or hard to go to the previous stage). My question is: do you gain something by having the 3 provider names in a database table?

FiveRockingFingers/FRF.DataAccess/EntityModels/Provider.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.ComponentModel.DataAnnotations;
+
+namespace FRF.DataAccess.EntityModels
+{
+    public class Provider
+    {
+        [Key]
+        public int Id { get; set; }
+        public string Name { get; set; }
+    }
```
gonzalorom, line 10: The convention for the project in a one-to-many relationship is to include navigation property at both ends. So here there is one navigator missed `public ICollection<ArtifactType> ArtifactType`. _Does it work as it is now?_ ***Yes***, is just to follow the convention.

gonzalorom (approved):
I like how it was implemented.

martinmena:
> Changes looks good but I would like to leave a comment. I bet you have been thinking hard on this and I believe this is an important decision with no return (or hard to go to the previous stage). My question is: do you gain something by having the 3 provider names in a database table?

I think there is one big advantage in having the provider names in the database, and that is that every artifact has a provider associated. This means that if in the future we have 5/6 different providers, and for example you want to create a project based on one of them, this relation allows you to filter every artifact by provider, instead of searching through an endless list. The other point is that multiple provides might have services with the same name (with this one I'm just speculating) and this would be a way to identify which is which.
Now, I don't know if the providers "name" is actually needed, but some kind of relation is useful in my opinion, and the only other option that I can think of right now is that we store only a number in the ArtifactType (working as a ProviderId but without the table) and then we map that number to the correspondent provider in the BackEnd (I personally don't like this approach, it's like "simulating" a table in the backend, and not in the database).

FiveRockingFingers/FRF.DataAccess/EntityModels/Provider.cs
martinmena: Good point! This is working right now, but I add it to follow the same convention.

matiasbeckerle:
The "simulation" you mention is the solution I was thinking about. I don't see it that bad as you see it 😁 -- anyways I'm OK with whatever you decide doing.

gonzalorom:
In my opinion, I like the current approach better. I personally feel that we would be adding more lines of code, enums (or whatever class we use), which can be solved by having these values in the database 🤔.

PR #100, opened 2021-02-10
[FIVE-300] Modificar métodos Assert..Equals en test donde esten los parámetros estén invertidos.

gonzalorom:
### Descripción: 
Teniendo en cuenta el método `Assert.Equal<T>(T expected, T actual)` correspondiente a la herramienta XUnit, la cual es utilizada en este proyecto para realizar Unit Test, donde:
***expected: El valor esperado***
***actual: El valor a comparar***
Se encontraron que varias implementaciones no respetaban dicha estructura.

### Tareas Realizadas:
- Se modificaron test donde la estructura (Expected , Actual) de los métodos Assert.Equals estaban invertidas.

### DIFF
--- a/test/FRF.Core.Tests/Services/ArtifactsServiceTests.cs
+++ b/test/FRF.Core.Tests/Services/ArtifactsServiceTests.cs
@@ -756,7 +756,7 @@ namespace FRF.Core.Tests.Services
             Assert.IsType<ServiceResponse<IList<CoreModels.ArtifactsRelation>>>(response);
             Assert.NotNull(response.Error);
             Assert.Equal(ErrorCodes.RelationAlreadyExisted, response.Error.Code);
-
+            Assert.Null(response.Value);
             Assert.Equal(artifactsRelationToSave[0].Artifact1Property, artifactsRelationInDb[0].Artifact1Property);
             Assert.Equal(artifactsRelationToSave[0].Artifact2Property, artifactsRelationInDb[0].Artifact2Property);
             Assert.Equal(artifactsRelationToSave[0].Artifact2Id, artifactsRelationInDb[0].Artifact2Id);
@@ -798,7 +798,7 @@ namespace FRF.Core.Tests.Services
             Assert.IsType<ServiceResponse<IList<CoreModels.ArtifactsRelation>>>(response);
             Assert.NotNull(response.Error);
             Assert.Equal(response.Error.Code, ErrorCodes.ArtifactNotExists);
-
+            Assert.Null(response.Value);
             Assert.Equal(artifactsRelationUpdated[0].Artifact1Property, artifactsRelationInDb[0].Artifact1Property);
             Assert.Equal(artifactsRelationUpdated[0].Artifact2Property, artifactsRelationInDb[0].Artifact2Property);
             Assert.Equal(artifactsRelationUpdated[0].Artifact2Id, artifactsRelationInDb[0].Artifact2Id);
--- a/test/FRF.Core.Tests/Services/ArtifactsServiceTests.cs
+++ b/test/FRF.Core.Tests/Services/ArtifactsServiceTests.cs
@@ -799,10 +799,10 @@ namespace FRF.Core.Tests.Services
             Assert.NotNull(response.Error);
             Assert.Equal(response.Error.Code, ErrorCodes.ArtifactNotExists);
 
-            Assert.Equal(artifactsRelationInDb[0].Artifact1Property, artifactsRelationUpdated[0].Artifact1Property);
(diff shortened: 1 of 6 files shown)


test/FRF.Core.Tests/Services/ArtifactsServiceTests.cs
```diff
@@ -755,12 +755,12 @@ public async Task SetRelationAsync_ReturnsNull_WhenIsAnyArtifactRepeated()
...
+            Assert.Equal(ErrorCodes.RelationAlreadyExisted, response.Error.Code);
 
-            Assert.Equal(artifactsRelationInDb[0].Artifact1Property, artifactsRelationToSave[0].Artifact1Property);
-            Assert.Equal(artifactsRelationInDb[0].Artifact2Property, artifactsRelationToSave[0].Artifact2Property);
-            Assert.Equal(artifactsRelationInDb[0].Artifact2Id, artifactsRelationToSave[0].Artifact2Id);
-            Assert.Equal(artifactsRelationInDb[0].Artifact1Id, artifactsRelationToSave[0].Artifact1Id);
+            Assert.Equal(artifactsRelationToSave[0].Artifact1Property, artifactsRelationInDb[0].Artifact1Property);
+            Assert.Equal(artifactsRelationToSave[0].Artifact2Property, artifactsRelationInDb[0].Artifact2Property);
+            Assert.Equal(artifactsRelationToSave[0].Artifact2Id, artifactsRelationInDb[0].Artifact2Id);
+            Assert.Equal(artifactsRelationToSave[0].Artifact1Id, artifactsRelationInDb[0].Artifact1Id);
```
martinmena, line 763: Is this needed in this test? I know it might not be exactly what this PR is for, but reading the test again, this two lists are created before the call to SetRelationAsync, so they're not checking if the response from the method is the correct one, they are basically verifying that the mapper works.
So my question is, shouldn't this be more like:
`Assert.Equal(artifactsRelationToSave[0].Artifact1Property, result.Value[0].Artifact1Property);` 
or something like that? And the same happens in the test below this one

martinmena (approved):

gperelstein (approved):

test/FRF.Core.Tests/Services/ArtifactsServiceTests.cs
gonzalorom: In both situations the main object of the test is to verify if returned values are an `Error.Code` when something happen. So like you said this equality is only verifying the correct work of the mapper, what in my opinion is not that bad. 
TL;DR: what we should check is if ***result.Value is equal to null*** because you are receiving a `response.Error.Code`.

gonzalorom: Although the scope of this ticket was only Assert.Equal, if you don't mind, I can add these validations 😄

martinmena: > In both situations the main object of the test is to verify if returned values are an Error.Code when something happen

If what you want to check is if the operation was a success or not, there is a boolean inside ServiceResponse that can tell you that. If ServiceResponse.Success is _true_ then there is going to be a Value and no Error, but if it is _false_ there is going to be an Error and no Value.

> So like you said this equality is only verifying the correct work of the mapper, what in my opinion is not that bad

Sound fair, but I would also check if the response from the method is the correct one, and that way, that the relations are created the right way (or updated in the other method). In my opinion, and since the method returns the created relations, it would be best to check each of them to make sure they are created the right way. If we don't, there is a chance that the method could change some of the properties of the relation without us knowing it (the artifacts, their selected setting, the relation type...). I'm not saying this method actually changes the relation, but this kind of tests allow us to verify they are working correctly in every way.

> Although the scope of this ticket was only Assert.Equal, if you don't mind guys, I can add these validations 😄

I have no problem with that, but I'll leave it up to you. If you think this changes deserve a new PR, I'm ok with that also, but I have no preference.

gonzalorom: > ...since the method returns the created relations, it would be best to check each of them to make sure they are created the right way.
>
Here is the problem, please check line 758. Like I said:

> TL;DR: what we should check is if result.Value is equal to null because you are receiving a response.Error.Code.
>
For that reason we can not check each of them because ***response.Value is equals to null***.

martinmena: Ohhhhh my bad, you're right, sorry. :man_facepalming:

gonzalorom: No worries mate, all good!!

PR #103, opened 2021-02-11
[FIVE-310] (BackEnd) Modificar formulario de creación de artefactos para tener como opción solo los tipos de artefactos correspondientes al Proveedor

martinmena:
## Descripción
En la creación de artefactos no se filtran los tipos de artefactos correspondientes a el proveedor seleccionado, mostrándose el listado completo, con lo que el proveedor que se seleccione en la primer pestaña no tienen ningún peso (el proveedor asociado a cada artefacto se obtiene de el tipo de artefacto)
Para corregir esto se debe crear un servicio de ArtifactTypes y su controller. En el debe existir un método para obtener todos los tipos de artefactos filtrados por un proveedor, y este listado se debe mostrar en el formulario de creación de artefactos luego de que se haya seleccionado un proveedor.

## Cambios realizados
En este PR se encuentran los cambios realizados sobre el BackEnd, que incluyen la creación de el servicio de ArtifactTypes, su controller y los tests de ambos.
Los cambios sobre el FrontEnd se encuentran en el PR #104 .

### DIFF
--- a/FiveRockingFingers/FRF.Core/Services/ArtifactTypesService.cs
+++ b/FiveRockingFingers/FRF.Core/Services/ArtifactTypesService.cs
@@ -1,0 +1,49 @@
+﻿using AutoMapper;
+using FRF.Core.Models;
+using FRF.Core.Response;
+using FRF.DataAccess;
+using Microsoft.EntityFrameworkCore;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace FRF.Core.Services
+{
+    public class ArtifactTypesService : IArtifactTypesService
+    {
+        private readonly DataAccessContext _dataContext;
+        private readonly IMapper _mapper;
+
+        public ArtifactTypesService(DataAccessContext dataContext, IMapper mapper)
+        {
+            _dataContext = dataContext;
+            _mapper = mapper;
+        }
+
+        public async Task<ServiceResponse<List<ArtifactType>>> GetAll()
+        {
+            var artifactTypes = await _dataContext.ArtifactType
+                .Include(at => at.Provider)
+                .ToListAsync();
+
+            var mappedArtifactTypes = _mapper.Map<List<ArtifactType>>(artifactTypes);
+            return new ServiceResponse<List<ArtifactType>>(mappedArtifactTypes);
+        }
+
+        public async Task<ServiceResponse<List<ArtifactType>>> GetAllByProvider(string providerName)
+        {
+            if (!await _dataContext.Providers.AnyAsync(p => p.Name == providerName))
+            {
+                return new ServiceResponse<List<ArtifactType>>(new Error(ErrorCodes.ProviderNotExists, $"There is no '{providerName}' provider"));
+            }
+
+            var artifactTypes = await _dataContext.ArtifactType
+                .Include(at => at.Provider)
+                .Where(at => at.Provider.Name == providerName)
+                .ToListAsync();
+
+            var mappedArtifactTypes = _mapper.Map<List<ArtifactType>>(artifactTypes);
+            return new ServiceResponse<List<ArtifactType>>(mappedArtifactTypes);
+        }
+    }
+}
--- a/FiveRockingFingers/FRF.Core/Services/ArtifactTypesService.cs
+++ b/FiveRockingFingers/FRF.Core/Services/ArtifactTypesService.cs
@@ -20,7 +20,7 @@ namespace FRF.Core.Services
             _mapper = mapper;
         }
 
-        public async Task<ServiceResponse<List<ArtifactType>>> GetAll()
+        public async Task<ServiceResponse<List<ArtifactType>>> GetAllAsync()
         {
             var artifactTypes = await _dataContext.ArtifactType
                 .Include(at => at.Provider)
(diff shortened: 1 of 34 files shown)


FiveRockingFingers/FRF.Core/Services/IArtifactTypesService.cs
```diff
@@ -0,0 +1,13 @@
...
+using FRF.Core.Response;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace FRF.Core.Services
+{
+    public interface IArtifactTypesService
+    {
+        Task<ServiceResponse<List<ArtifactType>>> GetAll();
+        Task<ServiceResponse<List<ArtifactType>>> GetAllByProvider(string providerName);
```
matiasbeckerle, line 11: Remember to always use the async suffix.

FiveRockingFingers/FRF.Web/Controllers/ArtifactTypesController.cs
```diff
@@ -0,0 +1,43 @@
...
+            var artifactTypes = await _artifactTypesService.GetAll();
+            var artifactTypesDto = _mapper.Map<IEnumerable<ArtifactTypeDTO>>(artifactTypes.Value);
+
+            return Ok(artifactTypesDto);
+        }
+
+        [HttpGet("{providerName}")]
+        public async Task<IActionResult> GetAllByProviderAsync(string providerName)
+        {
+            var artifactTypes = await _artifactTypesService.GetAllByProvider(providerName);
```
matiasbeckerle, line 37: Don't you have to take into account 404 when the provider doesn't exist?

test/FRF.Web.Tests/Controllers/ArtifactTypesControllerTests.cs
```diff
@@ -0,0 +1,123 @@
...
+            var resultValue = Assert.IsAssignableFrom<IEnumerable<ArtifactTypeDTO>>(okResult.Value);
+            var resultValueList = resultValue.ToList();
+
+            AssertCompareList(artifactTypes, resultValueList);
+
+            _artifactTypesService.Verify(mock => mock.GetAll(), Times.Once);
+        }
+
+        [Fact]
+        public async Task GetAllByProviderAsync_ReturnsOk()
```
matiasbeckerle, line 97: And when the provider doesn't exist?

matiasbeckerle (approved):
Code looks good but I believe you should consider 1 other case.

gperelstein (approved):

FiveRockingFingers/FRF.Core/Services/ArtifactTypesService.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            var artifactTypes = await _dataContext.ArtifactType
+                .Include(at => at.Provider)
+                .ToListAsync();
+
+            var mappedArtifactTypes = _mapper.Map<List<ArtifactType>>(artifactTypes);
+            return new ServiceResponse<List<ArtifactType>>(mappedArtifactTypes);
+        }
+
+        public async Task<ServiceResponse<List<ArtifactType>>> GetAllByProvider(string providerName)
```
gonzalorom, line 33: Maybe some trim here.

FiveRockingFingers/FRF.Core/Services/ArtifactTypesService.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            if (!await _dataContext.Providers.AnyAsync(p => p.Name == providerName))
+            {
+                return new ServiceResponse<List<ArtifactType>>(new Error(ErrorCodes.ProviderNotExists, $"There is no '{providerName}' provider"));
+            }
+
+            var artifactTypes = await _dataContext.ArtifactType
+                .Include(at => at.Provider)
+                .Where(at => at.Provider.Name == providerName)
+                .ToListAsync();
```
gonzalorom, line 43: Only for curiosity 💭. Why haven't you included the Artifact navigator as well?

gonzalorom (approved):

FiveRockingFingers/FRF.Core/Services/ArtifactTypesService.cs
martinmena: I'm sorry, but I'm not quite sure what you mean. Are you asking why isn't this query a bit bigger and also gets the information about the artifacts related to the artifactTypes? If that's so, it's because we don't need the artifacts (in this case we just want a list of artifactTypes according to a provider) and if we add that, we would be passing around a lot more information that wouldn't be used (at least in this case). I don't know when that information would be useful, but if we ever need it we could create a new method or modify this one.
Having said that, I don't know if it answers your question.

gonzalorom: Yes, that answers my question. Good approach to not bring unnecessary information. I was wondering that in case we ever need to obtain the artifacts of an artifact type. But as you say, when the time comes we can make a method in the place where it corresponds. Thanks for answering the question mate!

FiveRockingFingers/FRF.Core/Services/IArtifactTypesService.cs
martinmena: Done!

FiveRockingFingers/FRF.Web/Controllers/ArtifactTypesController.cs
martinmena: Great catch!

PR #105, opened 2021-02-11
[Five-302] (Models) Crear implementacion de precios para Artifact AWS EC2

gperelstein:
Se debe crear una clase que represente el servicio AmazonEC2 (la clase se llama AwsEc2) y que contenga un método capaz de calcular el precio final de un artefacto de este tipo. Esta clase hereda de la clase Artifact y todas las variables necesarias para realizar el cálculo de precio son tomadas de la propiedad Settings.

Además se debe crear un archivo xsd para que se validen las settings de un artefacto AmazonEC2 al momento de ser guardado o de ser actualizado y de esta forma evitar problemas en el futuro al querer usar el método GetPrice y no encontrar los nodos necesarios en el XML correspondiente a las settings.

### DIFF
--- a/FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
+++ b/FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
@@ -40,6 +40,10 @@ namespace FRF.Core
                 .ForMember(dest => dest.Id, act => act.Ignore())
                 .ForMember(dest => dest.ArtifactType, act => act.Ignore())
                 .ForMember(dest => dest.Project, act => act.Ignore());
+            CreateMap<DataAccess.EntityModels.Artifact, Models.AwsArtifacts.AwsEc2>()
+                .ForMember(dest => dest.Id, act => act.Ignore())
+                .ForMember(dest => dest.ArtifactType, act => act.Ignore())
+                .ForMember(dest => dest.Project, act => act.Ignore());
             CreateMap<DataAccess.EntityModels.ArtifactType, Models.ArtifactType>()
                 .ReverseMap();
             CreateMap<DataAccess.EntityModels.UsersByProject, Models.UsersProfile>()
--- a/FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
+++ b/FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
@@ -40,10 +40,7 @@ namespace FRF.Core
                 .ForMember(dest => dest.Id, act => act.Ignore())
                 .ForMember(dest => dest.ArtifactType, act => act.Ignore())
                 .ForMember(dest => dest.Project, act => act.Ignore());
-            CreateMap<DataAccess.EntityModels.Artifact, Models.AwsArtifacts.AwsEc2>()
-                .ForMember(dest => dest.Id, act => act.Ignore())
-                .ForMember(dest => dest.ArtifactType, act => act.Ignore())
-                .ForMember(dest => dest.Project, act => act.Ignore());
+            CreateMap<DataAccess.EntityModels.Artifact, Models.AwsArtifacts.AwsEc2>();
             CreateMap<DataAccess.EntityModels.ArtifactType, Models.ArtifactType>()
                 .ReverseMap();
             CreateMap<DataAccess.EntityModels.UsersByProject, Models.UsersProfile>()
--- a/FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsEc2.cs
+++ b/FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsEc2.cs
@@ -1,0 +1,325 @@
+﻿using FRF.Core.Response;
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace FRF.Core.Models.AwsArtifacts
+{
+    public class AwsEc2 : Artifact
+    {
+        public const decimal PartialStorageDiscount = 0.5m;
+
+        //Compute instance properties
+        public int HoursUsedPerMonth { get; set; }
+        public string PurchaseOption { get; set; }
+        public decimal InstancePricePerUnit0 { get; set; }
+        public decimal InstancePricePerUnit1 { get; set; }
+        public string LeaseContractLength { get; set; }
+
+        //EBS Storage properties
+        public string VolumenApiName { get; set; }
+
+        public int NumberOfGbStorageInEbs { get; set; }
+        public decimal EbsPricePerUnit { get; set; }
+
+        //EBS Snapshots properties
+        public int NumberOfGbChangedPerSnapshot {get; set; }
+        public decimal NumberOfSnapshotsPerMonth { get; set; }
+        public decimal SnapshotPricePerUnit { get; set; }
+
+        //EBS IOPS properties
+        public int NumberOfIops { get; set; }
+        public decimal IopsPricePerUnit { get; set; }
+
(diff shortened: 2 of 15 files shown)


FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsEc2.cs
```diff
@@ -0,0 +1,325 @@
...
+
+        private void GetComputeInstaceProperties()
+        {
+            HoursUsedPerMonth = int.Parse(Settings.Element("product0").Element("hoursUsedPerMonth").Value);
+
+            InstancePricePerUnit0 = decimal.Parse(Settings.Element("product0")
+                .Element("pricingDimensions")
+                .Element("range0")
+                .Element("pricePerUnit")
+                .Value, CultureInfo.InvariantCulture);
```
gonzalorom, line 219: There may be a possible bug here, maybe it's time to consider implementing some kind of decimal resolver.
Because, based on my experience with AWS pricing API, a price can come converted to scientific expression (Ex: **4e-7** or **5E-06**) and by using `decimal.Parse(Settings.Element("foo").Element("bar").Value, CultureInfo.InvariantCulture)` the application will crash 💥.  You can use some this like:
```csharp
private decimal GetDecimalPrice(string pricePerUnit)
        {
           //some validation//
            if (decimal.TryParse(pricePerUnit, NumberStyles.AllowExponent, CultureInfo.InvariantCulture,
                out decimal decimalPrice)) return decimalPrice;

            if (decimal.TryParse(pricePerUnit, NumberStyles.Currency, CultureInfo.InvariantCulture,
                out decimalPrice)) return decimalPrice;
           //If you reach here something went wrong, and you need to call AWS pricing API developers 🤷‍♂️//
            return -1;
        }
```
`NumberStyles.AllowExponent` do all the magic to converted scientific expression. And if the value is already a decimal : `NumberStyles.Currency, CultureInfo.InvariantCulture` does all the rest (**0,123** or **0.123**).

Please do not feel like something mandatory, I only talk about what I was finding while reviewing the API responses. Feel free to implement the solution that you like the most!

FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsEc2.cs
```diff
@@ -0,0 +1,325 @@
...
+        public int NumberOfMbpsThroughput { get; set; }
+        public decimal ThroughputPricePerUnit { get; set; }
+
+        //Data Transfer properties
+        public int NumberOfGbTransferIntraRegion { get; set; }
+        public decimal IntraTegionDataTransferPricePerUnit { get; set; }
+
+        override public decimal GetPrice()
+        {
+            GetAllProperties();
```
matiasbeckerle, line 46: This is a weird pattern. Shouldn't you load the whole class at construction time?

FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsEc2.cs
```diff
@@ -0,0 +1,325 @@
...
+        }
+
+        private decimal GetEbsPrice()
+        {
+            return NumberOfGbStorageInEbs * EbsPricePerUnit;
+        }
+
+        private decimal GetInstancePrice()
+        {
+            var price = -1m;
```
matiasbeckerle, line 133: -1?

FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsEc2.cs
```diff
@@ -0,0 +1,325 @@
...
+            if (PurchaseOption.Equals("No Upfront", StringComparison.InvariantCultureIgnoreCase) ||
+                PurchaseOption.Equals("NoUpfront", StringComparison.InvariantCultureIgnoreCase))
+            {
+                price = GetPriceNoUpfront();
+            }
+
+            return price;
+        }
+
+        private decimal GetPriceAllUpfront1Yr()
```
matiasbeckerle, line 176: You can use the whole word: "Year"

matiasbeckerle (approved):

FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsEc2.cs
gperelstein: I tried that but I found some problems with the unit tests. The method `GetAllProperties();` needs to access the settings property that is inherited from the artifact class). The problem appears in the tests when I initialize the class under test before arranging the settings. There is a proper way to solve this in the tests?

FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsEc2.cs
gperelstein: Good catch! I tested this and you are right. I implemented the method that you propose and created a new method to check if every price is correct

FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsEc2.cs
gperelstein: Good catch! It is an error. It was for some tests that I was doing. I already changed to 0

FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsEc2.cs
gperelstein: Done!

FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
```diff
@@ -40,6 +40,10 @@ public AutoMapperProfile()
                 .ForMember(dest => dest.Id, act => act.Ignore())
                 .ForMember(dest => dest.ArtifactType, act => act.Ignore())
                 .ForMember(dest => dest.Project, act => act.Ignore());
+            CreateMap<DataAccess.EntityModels.Artifact, Models.AwsArtifacts.AwsEc2>()
+                .ForMember(dest => dest.Id, act => act.Ignore())
+                .ForMember(dest => dest.ArtifactType, act => act.Ignore())
+                .ForMember(dest => dest.Project, act => act.Ignore());
```
martinmena, line 46: Sorry to bring it back to the table, but as we discussed in a previous PR, I don't think we should ignore this properties when getting the record from the database. That AwsEc2 model represents an Artifact that exists in the database, with an Id, an ArtifactType and a Project. I don't see the need to ignore them

martinmena (approved):

FiveRockingFingers/FRF.Core/AutoMapperProfile.cs
gperelstein: I didn't remember what we ended up agreeing on about this. But giving it a second think, maybe it would benefit not ignore them. I have deleted these ignoring sentences for now

FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsEc2.cs
matiasbeckerle: Yes.

```csharp
var ec2 = new AwsEc2
{
    YourSettingsXml = "...xml"
};
```

gperelstein: I already create getters for the properties of the AwsEc2 class and a constructor that receives the settings as a parameter

matiasbeckerle (approved):

gonzalorom (approved):

PR #110, opened 2021-02-18
[Five-324] Corregir métodos Save, Update y GetAllRelationsOfAnArtifactAsync en el servicio de Artefactos

gperelstein:
Se deben modificar los métodos Save y Update del servicio de artefactos para descomentar la sección que se encarga de la validación del xml correspondientes a las settings del artefacto que se quiere guardar o modificar. Para eso se debe buscar el tipo al cual corresponde el artefacto, al momento de realizar el Save o el Update, ya que la validación se realizara según el tipo de artefacto que sea.

El método GetAllRelationsOfAnArtifactAsync del servicio de artefactos se debe modificar para que devuelva un error en caso de que el artefacto, del cual se quieren obtener las relaciones, no exista.

### DIFF
--- a/FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
+++ b/FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
@@ -162,18 +162,19 @@ namespace FRF.Core.Services
                 return new ServiceResponse<Artifact>(new Error(ErrorCodes.ProjectNotExists, $"There is no project with Id = {artifact.ProjectId}"));
             }
 
-            if(! await _dataContext.ArtifactType.AnyAsync(at => at.Id == artifact.ArtifactTypeId))
+            var artifactType = await _dataContext.ArtifactType.Include(at => at.Provider).SingleOrDefaultAsync(at => at.Id == artifact.ArtifactTypeId);
+
+            if (artifactType == null)
             {
                 return new ServiceResponse<Artifact>(new Error(ErrorCodes.ArtifactTypeNotExists, $"There is no artifact type with Id = {artifact.ArtifactTypeId}"));
-            }            
+            }
 
-            //----------Uncomment after create service for artifact types--------------
-            /*
+            artifact.ArtifactType = _mapper.Map<ArtifactType>(artifactType);
+
             if (!_settingsValidator.ValidateSettings(artifact))
             {
                 return new ServiceResponse<Artifact>(new Error(ErrorCodes.InvalidArtifactSettings, $"Settings are invalid"));
             }
-            */
 
             // Maps the artifact into an EntityModel, deleting the Id if there was one, and setting the CreatedDate field
             var mappedArtifact = _mapper.Map<EntityModels.Artifact>(artifact);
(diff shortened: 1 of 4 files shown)


FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -164,18 +164,19 @@ public async Task<ServiceResponse<Artifact>> Save(Artifact artifact)
...
-            if(! await _dataContext.ArtifactType.AnyAsync(at => at.Id == artifact.ArtifactTypeId))
+            var artifactType = await _dataContext.ArtifactType.Include(at => at.Provider).SingleOrDefaultAsync(at => at.Id == artifact.ArtifactTypeId);
+
+            if (artifactType == null)
             {
                 return new ServiceResponse<Artifact>(new Error(ErrorCodes.ArtifactTypeNotExists, $"There is no artifact type with Id = {artifact.ArtifactTypeId}"));
-            }            
+            }
+
+            artifact.ArtifactType = _mapper.Map<ArtifactType>(artifactType);
```
martinmena, line 174: I'm not quite sure if this is needed or not, but I believe that having the ArtifactTypeId property associated to the correct Id should be enough, although as I said, I'm not sure. (If that's so, that means the previous query to the database could be a little simpler, and not include the Provider)

martinmena (approved):

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gperelstein: The problem is that  the settings validator uses a different file for XML validation, depending on the artifact type and I think we should check the provider first (as we did with the MapArtifact function in this service) just in case we have two artifact types with the same name and different provider

gperelstein: Giving it a second thought maybe we could change both switch (the one in the MapArtifact function and the one in the SettingsValidator) to just check for the ArtifactTypeId. What do you think?

matiasbeckerle: It works both ways. If you ask me? I think I prefer to avoid the query in line 167 and just let the db blow up if the provided _ArtifactTypeId_ doesn't exist. Which means... code as simple as possible until complexity is needed.

matiasbeckerle (approved):

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gperelstein: The problem is that we need that query for the settings validation. For now, the settings validation is doing something like:

 ```
switch (artifact.ArtifactType.Provider.Name)
{
    case ArtifactTypes.Custom:
        "Validate Settings with file CustomArtifact.xsd"
        break;
    case ArtifactTypes.Aws:
        switch (artifact.ArtifactType.Name)
        {
            case AwsS3Descriptions.Service:
                 "Validate Settings with file AwsS3.xsd"
                break;
            case AwsEc2Descriptions.Service:
                "Validate Settings with file AwsEc2.xsd"
                break;
        }
}
```
So we need the provider name and the artifact type name.

gonzalorom (approved):

PR #112, opened 2021-02-25
[Five-330] (DAL) Añadir edición de artefactos custom

gperelstein:
Se debe añadir la posibilidad de editar artefactos Custom. Para ello se deben borrar las relaciones de las settings de las cuales se haya cambiado el nombre. Es por esto que en este PR se incorpora un método al servicio de artefactos que me permita eliminar varias relaciones a la vez

### DIFF
--- a/FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
+++ b/FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
@@ -241,7 +241,11 @@ namespace FRF.Core.Services
                 return new ServiceResponse<Artifact>(new Error(ErrorCodes.InvalidArtifactSettings, $"Settings are invalid"));
             }
 
-            await DeleteRelationsOfUpdatedArtifact(artifact.Id, artifact.Settings, result.Settings);
+            var artifactsRelationsIdsToDelete = await GetRelationsToDeleteOfUpdatedArtifact(artifact.Id, artifact.Settings, result.Settings);
+
+            var artifactsRelationsToDelete = await _dataContext.ArtifactsRelation.Where(ar => artifactsRelationsIdsToDelete.Contains(ar.Id)).ToListAsync();
+
+            _dataContext.ArtifactsRelation.RemoveRange(artifactsRelationsToDelete);
 
             //Updates the artifact
             result.Name = artifact.Name;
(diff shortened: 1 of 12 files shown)


FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -348,6 +348,27 @@ public async Task<ServiceResponse<ArtifactsRelation>> DeleteRelationAsync(Guid a
...
+            {
+                var artifactsRelation = await _dataContext.ArtifactsRelation
+                .FirstOrDefaultAsync(ar => ar.Id.Equals(artifactRelationId));
+                if (artifactsRelation == null)
+                {
+                    return new ServiceResponse<IList<ArtifactsRelation>>(new Error(ErrorCodes.RelationNotExists, $"There is no relation with Id={artifactRelationId}"));
+                }
+                artifactsRelations.Add(artifactsRelation);
+                _dataContext.ArtifactsRelation.Remove(artifactsRelation);
+            }
```
gonzalorom, line 365: Is there a way where we can accomplish this without so many requests to the database? In which, that responsibility can be delegated to the DB.

gperelstein: Good observation, I changed the multiples calls to the Remove method for only one call to the RemoveRange method

gonzalorom: Excellent first step mate, thanks for considering this. I still think we can go further on to reduce the queries to the DB. This will lead us to decide whether we want to return the id that does not exist.

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -348,6 +348,27 @@ public async Task<ServiceResponse<ArtifactsRelation>> DeleteRelationAsync(Guid a
...
         }
 
+        public async Task<ServiceResponse<IList<ArtifactsRelation>>> DeleteRelationsAsync(IList<Guid> artifactRelationIds)
+        {
+            var artifactsRelations = new List<EntityModels.ArtifactsRelation>();
+
+            foreach (var artifactRelationId in artifactRelationIds)
+            {
+                var artifactsRelation = await _dataContext.ArtifactsRelation
+                .FirstOrDefaultAsync(ar => ar.Id.Equals(artifactRelationId));
```
matiasbeckerle, line 358: This isn't performant. You should go fetch all of them on a single db trip.

```
var artifactsRelationshipsToDelete = await _dataContext.ArtifactsRelation
  .Where(ar => artifactRelationIds.Contains(ar.Id))
  .ToList()
```

gonzalorom: That was what I wanted to get to haha.

gperelstein: Excellent, I understand now. I already did the changes

matiasbeckerle (approved):

gonzalorom (approved):
Just if you want to return the non-existent relations, you can do something like this :
```csharp
var relationsIdFromDb = await _dataContext.ArtifactsRelation.Select(ar => ar.Id).ToListAsync();
var relationsThatDoNotExist = artifactRelationIds.Except(relationsIdFromDb).ToList();
if (relationsThatDoNotExist.Any())
{
    return new ServiceResponse<IList<ArtifactsRelation>>(new Error(ErrorCodes.RelationNotExists,
        $"There is no relation with Id={string.Join(", Id=", relationsThatDoNotExist)}"));
}
```

martinmena (approved):

gperelstein:
> Just if you want to return the non-existent relations, you can do something like this :
> 
> ```cs
> var relationsIdFromDb = await _dataContext.ArtifactsRelation.Select(ar => ar.Id).ToListAsync();
> var relationsThatDoNotExist = artifactRelationIds.Except(relationsIdFromDb).ToList();
> if (relationsThatDoNotExist.Any())
> {
>     return new ServiceResponse<IList<ArtifactsRelation>>(new Error(ErrorCodes.RelationNotExists,
>         $"There is no relation with Id={string.Join(", Id=", relationsThatDoNotExist)}"));
> }
> ```

Greate I idea! Already implemented!

PR #114, opened 2021-02-25
[FIVE-327] Agregar filtros de proyectos por categoria

gonzalorom:
### Cambios realizados:
FIVE-327(3)

- Se creó componente NavMenuFilter el cual filtra los proyectos por categoría.
- Se agregó hooks dentro NavMenu para gestionar los estados de los proyectos filtrados y los filtros seleccionados.

### DIFF
--- a/FiveRockingFingers/FRF.Web/ClientApp/src/commons/NavMenu.tsx
+++ b/FiveRockingFingers/FRF.Web/ClientApp/src/commons/NavMenu.tsx
@@ -1,42 +1,133 @@
 import * as React from 'react';
-import { Menu, ProSidebar, SidebarContent, SidebarFooter } from 'react-pro-sidebar';
+import { Menu, ProSidebar, SidebarContent, SidebarFooter, SubMenu } from 'react-pro-sidebar';
 import 'react-pro-sidebar/dist/css/styles.css';
 import { Link } from 'react-router-dom';
 import Project from '../interfaces/Project';
+import Category from "../interfaces/Category";
 import '../styles/NavMenu.css';
 import NavMenuItem from './NavMenuItem';
+import NavMenuFilter from "./NavMenuFilter";
 import ProjectService from '../services/ProjectService';
+import CategoryService from "../services/CategoryService";
+import TuneTwoToneIcon from "@material-ui/icons/TuneTwoTone";
+import { Theme, withStyles, createStyles } from "@material-ui/core/styles";
+import Badge from "@material-ui/core/Badge";
+import {green} from "@material-ui/core/colors";
+import { Typography } from '@material-ui/core';
+
+const StyledBadge = withStyles((theme: Theme) =>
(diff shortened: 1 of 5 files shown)


gperelstein (approved):
It looks amazing!

FiveRockingFingers/FRF.Web/ClientApp/src/commons/NavMenu.tsx
```diff
@@ -1,42 +1,133 @@
...
+    setProjectsFiltered([]);
+  }
+
+  const handleNavMenuItem = () => {
+    return projectsFiltered.length === 0 && categoriesFilter.length === 0
+      ? projects.map((project) => (
+          <NavMenuItem key={project.id} project={project} />
+        ))
+      :  
+      projectsFiltered.length === 0 ?
```
matiasbeckerle, line 88: This is odd -- here you're using the same condition as on the beginning. Are you sure this works as you intended?

gonzalorom: Thank you for the review. Yes its working as is intended and this is why: There are three arrays: 
1. `projects` projects from the API
2. `categoriesFilter` all the categories selected to use as filters.
3. `projectsFiltered` Projects that meet the filters in the previous array.

If you chose any filter it will go inside `categoriesFilter` and if there is any project with that category it goes to `projectsFiltered`. So here first check if `categoriesFilter` and `projectsFiltered` are empty, if this is true, you didn't choose any filter, and it has to render all the projects from `projects`. If that is false, and you have chosen at least one filter, but there is not any projects with that condition,  `projectsFiltered` will be empty, and it has to render a message ***"Categoria sin proyectos"***. For the last condition, if all that was false, that's means you have at least one project that meets the category filter.

matiasbeckerle: It's very hard to read hehehe -- maybe there's a more explicit or verbose way to code this? Maybe use named variables that represents something? Like `var hasProjects = conditionA && conditionB;`  or something that makes sense in this context?

gonzalorom: You're right, that would be easier going to read.

martinmena (approved):
Nice one, looks really clean!
Just one question that I realized watching the Gif, is it correct that when you filter by two categories, you don't get a list of projects with the two categories associated, but instead a list of every project that has at least one of them. Is that the intended behavior? Again, just a question, no need to change anything if you think it's correct as it is.

FiveRockingFingers/FRF.Web/ClientApp/src/commons/NavMenu.tsx
gonzalorom: Done.

PR #117, opened 2021-03-02
[FIVE-345] Control de ciclo infinito de actualización por relaciones

martinmena:
## Descripción
Se debe controlar que no se hagan referencias circulares entre settings para evitar entrar en ciclos de actualizaciones infinitas, ya que al cambiar el valor de una setting actualiza el valor de las settings a las que hace referencia. En la imagen se muestran distintas alternativas para que esto ocurra (por supuesto que hay muchísimas más).
image

Siguiendo el primer ejemplo, como se actualiza S1 en el artefacto 1, y esta tiene una referencia hacia la S1 del artefacto 2, esta también tiene que actualizarse. Cuando se actualiza la setting del artefacto 2, como esta tiene una referencia hacia la S1 del artefacto 1, esta tiene que actualizarse nuevamente, lo que dispararía una nueva actualización sobre la S1 del artefacto 2, y así sucesivamente. 

## Cambios realizados
Para esto se optó por trabajar las relaciones de los proyectos como un digrafo (grafo dirigido) tomando como nodos los pares [artifactId, setting name] y como las trazas o caminos a las propias relaciones. De esta manera, usando un algoritmo de búsqueda (en este caso DFS) se puede detectar la presencia de ciclos, cubriendo así todos los casos.
En los tests de los métodos SetRelationAsync y UpdateRelationAsync se crearon dos métodos para testear esta funcionalidad, en ambos casos probando que se detecten los ciclos mencionados en los últimos dos ejemplos de la imagen. (El primer ejemplo se detecta como relación duplicada, ya que se relacionan dos veces a los mismos artefactos)

Además de esto, se modificaron algunos de los tests ya existentes para crear los artefactos y sus relaciones en la base de datos de test, ya que por como estaba creado antes, traía varios problemas con dicha detección de ciclos.

### DIFF
--- a/FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
+++ b/FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
@@ -82,7 +82,7 @@ namespace FRF.Core.Services
         /// <param name="artifactsRelations"></param>
         /// <param name="isAnUpdate">True if is to update, False if is a set</param>
         /// <returns>True if At least one of the artifact relation provided already exist </returns>
-        private bool IsAnyRelationRepeated(IList<EntityModels.ArtifactsRelation> dbArtifactRelations,
+        private bool IsAnyRelationRepeated(IList<ArtifactsRelation> dbArtifactRelations,
             IList<ArtifactsRelation> artifactsRelations, bool isAnUpdate)
         {
             var relations = new List<ArtifactsRelation>(artifactsRelations);
@@ -316,18 +316,20 @@ namespace FRF.Core.Services
             var artifactsExist = await DoArtifactsExist(artifactRelations);
             if (artifactsExist) return new ServiceResponse<IList<ArtifactsRelation>>(new Error(ErrorCodes.ArtifactNotExists, "At least one of the artifact Ids provided doesn't exist"));
 
-            var dbArtifactRelations = await _dataContext.ArtifactsRelation.Where(ar =>
+            var artifactResponse = await Get(artifactRelations[0].Artifact1Id);
+            var projectRelationsResponse = await GetAllRelationsByProjectIdAsync(artifactResponse.Value.ProjectId);
+
+            var dbArtifactRelations = projectRelationsResponse.Value.Where(ar =>
                     ar.Artifact1Id == artifactRelations[0].Artifact1Id ||
                     ar.Artifact2Id == artifactRelations[0].Artifact2Id ||
                     ar.Artifact2Id == artifactRelations[0].Artifact1Id ||
-                    ar.Artifact1Id == artifactRelations[0].Artifact2Id)
-                .ToListAsync();
+                    ar.Artifact1Id == artifactRelations[0].Artifact2Id).ToList();
 
             var relationsRepeated = IsAnyRelationRepeated(dbArtifactRelations, artifactRelations, isAnUpdate: false);
             if (relationsRepeated)
                 return new ServiceResponse<IList<ArtifactsRelation>>(new Error(ErrorCodes.RelationAlreadyExisted, "At least one of the relations already existed"));
 
-            var cycleDetected = await IsCircularReferenceAsync(artifactRelations, false);
+            var cycleDetected = IsCircularReference(dbArtifactRelations, artifactRelations, false);
             if (cycleDetected)
                 return new ServiceResponse<IList<ArtifactsRelation>>(new Error(ErrorCodes.RelationCycleDetected, "These relations would generate at least one cycle"));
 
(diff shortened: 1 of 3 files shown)


FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -396,6 +404,118 @@ public async Task<ServiceResponse<ArtifactsRelation>> DeleteRelationAsync(Guid a
             await _dataContext.SaveChangesAsync();
 
             return new ServiceResponse<IList<ArtifactsRelation>>(artifactsRelationsNew);
-        }        
+        }
+
+        private async Task<bool> IsCircularReferenceAsync(IList<ArtifactsRelation> newRelations, bool isUpdate)
```
matiasbeckerle, line 409: I'm sorry but I don't understand why you need this code, neither why you need to tell consumers of your API that you detected a circular reference.

I would like to know more: where is the error happening? On a save with EF?

martinmena: So, in the application, inside each project you can create multiple artifacts and relations between them. The idea of the artifacts is that they represent a service from a provider (at the moment they can only be an S3 and an EC2 from AWS, but many more in the future) or a custom artifact, which can store some custom values to be used later. The relations are connections between any artifacts that let them reference one value from the other. This is useful when anyone is creating a project because let's say you have 5 API Gateways in Amazon, and each one is expected to receive 500 requests per minute, but in the future that value can go up to 700. You could create the 5 artifacts representing the API Gateways, and another custom artifact with a setting "amount of requests"=500 (settings are key-value pairs). In the future, when you need to estimate the price of this API Gateways, you could change the one value in the custom artifact, and affect every artifact that has a reference to it. Another important note about this relations is that they go in one direction (no the boy band), and that tells the system which one is the value, and which one is the one with a reference to it.

When we where discussing how this relations where going to take part in the price calculations (we need the final value of each setting to calculate the price of each service), after taking into account a few alternatives, we ended up deciding that we won't store references in the database, to avoid having trouble when we get the final price for a project, but instead we're storing the actual value of the setting referenced in every artifact. So, continuing the same example, when we go to the database and retrieve one of the API Gateways stored, in the setting "requests" (the one in the API Gateway, not the custom artifact. In this case there would be a relation between the two settings. From the custom artifact, setting "amount of requests", **to** the API Gateway artifact, setting "requests") there would also be a 500. If the value in the custom artifact changes, the setting in the API Gateway needs to be updated also, changing from 500 to 700, for example. This is where we need the cycle control. (If you want to know more about this, and why we choose this approach instead of the others (or which are the other alternatives we considered) there is a word document uploaded to the ticket. If you want, I can send it to you via slack).

If we have, like in the first example in the PR description, two settings related in both ways, it means that when we update one of the values, the other has to be updated also. This creates an infinite loop (the first one updates, and so the second one has to update. The second one updates, and so the first one has to update...). There are multiple ways this can be achieved (the other two examples can help to illustrate), and that's the basic reason this method exists. Hope it makes this a little clearer. And sorry for the huge amount of text, I'll try to be a little more concise in the future.

TL;DR: Each artifact has multiple settings, and they can "take" the value from another setting in another artifact. If both of them are trying to take the value from each other, this creates a cycle that never ends and the system might explode :boom: (I'm just guessing here).

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -396,6 +404,118 @@ public async Task<ServiceResponse<ArtifactsRelation>> DeleteRelationAsync(Guid a
...
+                return false;
+            }
+
+            // Getting all the project relations. This might be modified in the future, to pass them via parameter.
+            // If this fails, there is no way to confirm there won't be any cycles, so the function returns true to avoid problems.
+            var projectId = _dataContext.Artifacts.Single(art => art.Id == newRelations[0].Artifact1Id).Project.Id;
+            var dbRelations = await GetAllRelationsByProjectIdAsync(projectId);
+            if (!dbRelations.Success)
+            {
+                return true;
```
gperelstein, line 426: I don't understand this. If you can't get all the relations for a project this function is going to return true and the `SetRelationAsync` and `UpdateRelationAsync` methods will return a service response with a RelationCycleDetected error, but that's not the problem here, right?

martinmena: Yes, you're correct. The thing here is that if you can't get all the relations in the project, you won't be able to determine if there is going to be a cycle when you introduce the new relations, and if you did save them, that could generate a pretty big problem in my opinion. That's why it's returning the error.
Now, about the type of error that the service response is going to return, I'm not sure which is the best. Of course, it's not a RelationCycleDetected (there wasn't a cycle detected), but if we want to change it, that means that we would have to return some other identifier or object to tell the parent method that the error happened while retrieving the project relations (as it is right now, there isn't a way to tell when or why the function returned true/false).
If you think it would be nice to have that, I'll get to work on it to return the same error that the _GetAllRelationsByProjectIdAsync_ would have returned (I think that would be the best alternative).

gperelstein: Maybe you can call GetAllRelationsByProjectIdAsync in the parent method and pass all the relations of the project to this method as a parameter? I don't know if that is the best alternative but that way you are going to know if there was a problem retrieving all the relations of a project before calling IsCircularReferenceAsync

martinmena: Yep, I think that's the best alternative, and that way we can also remove one call to the database when we get the relations of the artifacts to check if they are repeated or not. This way we would get them all together at the beginning, and avoid getting them once to check that, and another time as part of the project relations.
Thanks!

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
matiasbeckerle: Thank you for taking the time to explain! I also saw the screenshots of @gperelstein's work in progress for the other parts. I'm not exactly sure what sort of advice to provide you folks. I have two things in my mind right now:

- I can't spend time on your project, not at this level of involvement. These are the things were a deep dive is needed or required.
- Based on the code I saw on previous PRs, the code in this one with the circular reference topic and the incoming code from Gonza I have the feeling that something isn't entirely right. The 3 different codes I saw shows sings of that because are difficult to read/understand.

Both things collide each other because I can't do a valid proposition with the involvement I have in the project. I know, I'm really slow at understanding hard problems, but hey everyone has defects hehe. So maybe what you're doing it's totally fine and this is just my silly brain not understanding the whole picture.

In regards of all the C# code I ever saw I have never seen the need of doing this. I also use EF and have to sort out not-so-common table relationships so for that reason is that I suspect this is more complicated than what it should be. Maybe it's a design thing? Maybe there's something we are missing somewhere? I'm not sure, sorry. I know my opinion is not useful but maybe you can use this as a fuel to explore other ideas in the future?

I cannot approve/disapprove this PR and probably the upcoming one since I don't really understand what's happening in these lines. Please feel free to move forward, I know you will figure it out eventually. In the future, any step you can move in favor of simplifying the artifact-artifact relationship the code will definitively benefit.

gonzalorom (approved):

gperelstein (approved):

PR #120, opened 2021-03-09
[Five-344] Actualización de valores de las settings cuando cambian las relaciones (o valores referenciados)

gperelstein:
En este PR se trabaja con la actualización de las settings de un artefacto al momento de modificar otro artefacto o de crear, borrar o actualizar una relación.

Al momento de crear una relación se actualizará el valor de setting que esté al final de la relación como se muestra en la figura:

image

En el caso de que una setting esté al final de dos relaciones se sumaran los valores de las settings al principio de las relaciones:

image

Cuando se borra una relación se restará el valor de la setting, del principio de la relación, del valor de la setting del final de la relación, a menos que la relación que se borre sea la única. En este caso el valor que tomará la setting del final de la relación será el último valor que haya tenido. En la siguiente figura se ejemplifica lo mencionado anteriormente:

image

Al actualizar una relación se tendrá que tener en cuanta la relación que se elimina y la relación que se genera.

Al momento de modificar las settings de un artefacto, primero se debe verificar que no se esté modificando el valor de una setting al final de una relación. Luego se deben buscar todas las demás settings relacionadas con la setting modificada y actualizar todos los valores y partir de esto se deben modificar las settings relacionadas con las nuevas settings modificadas y continuar con el proceso hasta actualizar todas las settings.


FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -269,10 +269,18 @@ public async Task<ServiceResponse<Artifact>> Update(Artifact artifact)
...
 
-            var artifactsRelationsToDelete = await _dataContext.ArtifactsRelation.Where(ar => artifactsRelationsIdsToDelete.Contains(ar.Id)).ToListAsync();
+            var settingsWithValueUpdated = GetSettingsWithValueUpdated(artifact.Settings, result.Settings);
 
-            _dataContext.ArtifactsRelation.RemoveRange(artifactsRelationsToDelete);
+            foreach(var setting in settingsWithValueUpdated)
+            {
+                if(await IsSettingIsAtEndOfAnyRelation(artifact.Id, setting))
+                {
+                    return new ServiceResponse<Artifact>(new Error(ErrorCodes.InvalidArtifactSettings, $"You are trying to modify a setting at the of a relation"));
```
gonzalorom, line 281: ..at the **end** of a relation"

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -487,5 +567,260 @@ private async Task<List<Guid>> FindRelationsToDeleteOfUpdatedArtifact(int artifa
...
+
+        public bool IsASettingAtBeginningOfRelation(string propertyName, ArtifactsRelation relation)
+        {
+            return (relation.RelationTypeId == 0 && relation.Artifact1Property == propertyName) || (relation.RelationTypeId == 1 && relation.Artifact2Property == propertyName);
+        }
+
+        public bool IsASettingAtEndOfRelation(string propertyName, ArtifactsRelation relation)
+        {
+            return (relation.RelationTypeId == 0 && relation.Artifact2Property == propertyName) || (relation.RelationTypeId == 1 && relation.Artifact1Property == propertyName);
+        }
```
gonzalorom, line 732: I understand that achieving an update of two Artifacts where their settings are within an XML and business rules such as those named at the beginning of the PR must be respected, is not an easy task at all. Perfect work and excellent logical thinking of the solution.

What you did works and meets the acceptance criteria, only that the code became somewhat difficult to follow, since there are many functions that perform the same thing, but have different names.

***Isn't it better to have descriptive function names?***

Yes and I am a big fan of **ROC** (Really Obvious Code), but I feel that in this case we doubled the number of functions and navigating all the code became difficult.

I am not saying that you run to modify it (since I would not have a recommendation on how to do it either), I am only saying the feeling it gave me when I read it.

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -487,5 +567,260 @@ private async Task<List<Guid>> FindRelationsToDeleteOfUpdatedArtifact(int artifa
...
+                return new ServiceResponse<IList<ArtifactsRelation>>(new Error(ErrorCodes.ArtifactNotExists, $"There is no artifact with Id = {artifactId}"));
+            }
+
+            var result = new List<EntityModels.ArtifactsRelation>();
+
+            if (isAManipulationOfRelation)
+            {
+                result = _dataContext.ArtifactsRelation.Local
+                    .Where(ar => ((ar.Artifact1Id == artifactId && ar.Artifact1Property.Equals(propertyName)) || ((ar.Artifact2Id == artifactId) &&
+                    ar.Artifact2Property.Equals(propertyName)))).ToList();
```
martinmena, line 613: I'm sorry but what is _Local_ used for? I've never seen it

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -487,5 +567,260 @@ private async Task<List<Guid>> FindRelationsToDeleteOfUpdatedArtifact(int artifa
...
+            }
+            else
+            {
+                result = await _dataContext.ArtifactsRelation.Include(ar => ar.Artifact1)
+                    .Include(ar => ar.Artifact2)
+                    .Where(ar => ((ar.Artifact1Id == artifactId && ar.Artifact1Property.Equals(propertyName)) || ((ar.Artifact2Id == artifactId) &&
+                    ar.Artifact2Property.Equals(propertyName)))).ToListAsync();
```
martinmena, line 620: Do you need to include Artifact1 and Artifact2 here, but not if isAManipulationOfRelation is true? And why would you set _isAManipulationOfRelation_ to true or false? As far as I can tell by the rest of the code, it's going to be false always

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -487,5 +567,260 @@ private async Task<List<Guid>> FindRelationsToDeleteOfUpdatedArtifact(int artifa
...
+
+                if (float.TryParse(artifactAtTheBegin.Settings.Element(settingAtTheBegining).Value, out float valueToUpdate))
+                {
+                    finalValueOfSetting += valueToUpdate;
+                }
+                else
+                {
+                    allValuesFromSettingsAreFloat = false;
+                }
+            }
```
martinmena, line 682: If the first setting is not a float value, you're still going through all the other settings just to return false. Maybe you could just return false instead of using the boolean, that way you finish the for loop right when you find a setting that isn't a float, and you can also avoid the if below, and just update the setting value (if you reach that, it means you didn't return false, so every setting is float)

Or you could also make it a while cycle like in _IsSettingIsAtEndOfAnyRelation_

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -296,6 +320,23 @@ public async Task<ServiceResponse<Artifact>> Update(Artifact artifact)
             return new ServiceResponse<Artifact>(mappedArtifact);
         }
 
+        private async Task<ServiceResponse<Artifact>> AuxiliarUpdateArtifact(Artifact artifact)
+        {
+            //Gets the artifact associated to it from the database
+            var result = await _dataContext.Artifacts
+                .Include(a => a.ArtifactType)
+                    .ThenInclude(at => at.Provider)
+                .Include(a => a.Project)
```
martinmena, line 329: Do you need to include all of these?

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -487,5 +567,260 @@ private async Task<List<Guid>> FindRelationsToDeleteOfUpdatedArtifact(int artifa
...
+            {
+                artifactToUpdate.Settings.Element(settingToUpdateName).Value = finalValueOfSetting.ToString();
+            }
+            else
+            {
+                return false;
+            }
+
+            var artifactUpdatedResponse = await AuxiliarUpdateArtifact(artifactToUpdate);
+            var artifactUpdated = artifactUpdatedResponse.Value;
```
martinmena, line 694: I don't understand this two lines. At the beginning of this method, you got the _artifactToUpdate_, then you made some modifications to the settings, and here you're getting it again from the database, making the same changes on the settings, and setting the ModifiedDate to now. Isn't it easier to just set the modifiedDate on the artifactToUpdate?

artifactToUpdate.ModifiedDate = DateTime.Now;

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -487,5 +567,260 @@ private async Task<List<Guid>> FindRelationsToDeleteOfUpdatedArtifact(int artifa
...
+        private async Task<ServiceResponse<IList<ArtifactsRelation>>> GetRelationsToUpdateAsync(int artifactId, string propertyName)
+        {
+            var existArtifactId = await _dataContext.Artifacts.AnyAsync(a => a.Id == artifactId);
+            if (!existArtifactId)
+            {
+                return new ServiceResponse<IList<ArtifactsRelation>>(new Error(ErrorCodes.ArtifactNotExists, $"There is no artifact with Id = {artifactId}"));
+            }
+
+            var result = await _dataContext.ArtifactsRelation.Include(ar => ar.Artifact1)
+                .Include(ar => ar.Artifact2)
```
martinmena, line 579: Do you need to include Artifact1 and Artifact2= I don't think you're using them

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -487,5 +567,260 @@ private async Task<List<Guid>> FindRelationsToDeleteOfUpdatedArtifact(int artifa
...
+        private async Task<ServiceResponse<IList<ArtifactsRelation>>> GetRelationsByArtifactAndSetting(int artifactId, string propertyName)
+        {
+            var existArtifactId = await _dataContext.Artifacts.AnyAsync(a => a.Id == artifactId);
+            if (!existArtifactId)
+            {
+                return new ServiceResponse<IList<ArtifactsRelation>>(new Error(ErrorCodes.ArtifactNotExists, $"There is no artifact with Id = {artifactId}"));
+            }
+
+            var result = await _dataContext.ArtifactsRelation.Include(ar => ar.Artifact1)
+                .Include(ar => ar.Artifact2)
```
martinmena, line 647: Do you need to include Artifact1 and Artifact2= I don't think you're using them

test/FRF.Core.Tests/Services/ArtifactsServiceTests.cs
```diff
@@ -488,6 +525,83 @@ public async Task UpdateAsync_ReturnsArtifact()
...
+                Settings = new XElement("Settings", new XElement(setting1Name, updatedSetting1Value))
+            };
+
+            _settingsValidator.Setup(mock => mock.ValidateSettings(It.IsAny<CoreModels.Artifact>()))
+                .Returns(true);
+
+            // Act
+            var result = await _classUnderTest.Update(artifactToUpdate);
+            var artifact2Response = await _classUnderTest.Get(artifact2.Id);
+            var artifact3Response = await _classUnderTest.Get(artifact3.Id);
```
martinmena, line 573: I believe this is considered as testing both methods, Update and Get. If you wanted to test only the update, you could get the artifact2 and artifact3 using the _dataAccess. Just to comment tho, I don't know if this is wrong, so no need to change it unless someone else says so.

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gperelstein: Yes, I agree the code of this PR it's difficult to follow and @martinmena mentions the same thing. I'm sorry about that. This ticket was very tricky. When I was working on this, I thought that maybe having two functions called IsASettingAtBeginningOfRelation and IsASettingAtEndOfRelation will be more clear than having only one like IsASettingAtBeginningOfRelation and using **!**IsASettingAtBeginningOfRelation to checking if the setting was at the end of a relation. I will look at this and I will try to reduce the number of functions

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
```diff
@@ -487,5 +567,260 @@ private async Task<List<Guid>> FindRelationsToDeleteOfUpdatedArtifact(int artifa
...
+            {
+                var artifactAtBeginOfRelationResponse = await Get(relation.Artifact1Id);
+                var artifactAtBeginOfRelation = artifactAtBeginOfRelationResponse.Value;
+                return artifactAtBeginOfRelation;
+            }
+            else
+            {
+                var artifactAtBeginOfRelationResponse = await Get(relation.Artifact2Id);
+                var artifactAtBeginOfRelation = artifactAtBeginOfRelationResponse.Value;
+                return artifactAtBeginOfRelation;
```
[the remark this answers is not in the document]
gperelstein: Yes, it was a mistake. Already fixed! Thanks!

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gperelstein: Sorry, this was when I was trying to do all the modifications in the database at once. I implemented transactions for this as you recommend to me. I already delete the use of the local cache

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gperelstein: The same as above, I delete the use of isAManipulationOfRelation variable

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gperelstein: The problem is that I need to reflect these changes in the database. I don't use the normal Update method because if I do that I will enter the circle of updating the settings again.

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gperelstein: I already change transform this loop into a for loop with a false return

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gperelstein: I already delete this includes

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gperelstein: I already delete this includes

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gperelstein: I already delete this includes

test/FRF.Core.Tests/Services/ArtifactsServiceTests.cs
gperelstein: I deleted the call of the Get method from every test that not corresponds to the Get method. I get the artifacts from the _dataAccess as you suggested

FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
gperelstein: I have checked the functions again. I deleted _IsASettingAtBeginningOfRelation_. There is another function that you think I can delete?

martinmena (approved):

PR #126, opened 2021-03-12
[FIVE-342] (Backend) Definir settings relacionables de cada artefacto

martinmena:
## Tarea a realizar
Cada tipo de artefacto debe tener un listado de settings que pueden relacionarse, así como también que tipo de valor tienen o esperan cada una de ellas.

## Cambios realizados
Para resolver esto, se definió una nueva propiedad en el modelo de artefactos, RelationalFields. Este es un diccionario con el nombre de la setting a relacionar como clave, y como valor tiene el tipo que esta setting acepta (los dos tipos definidos por el momento se basan en los valores que aceptan los servicios S3 y EC2 de Aws).

El objetivo de esto es tener control sobre las asociaciones de valores que generen los usuarios, por un lado, para que al momento de realizar una relación entre artefactos, solo se puedan relacionar settings del mismo tipo. Es decir, que si una setting espera un valor Entero (por ejemplo la cantidad de requests que recibe al mes), no se pueda asociar una cadena o un decimal. Por otro lado, también se busca tener control sobre las settings que son relacionables de cada artefacto, para evitar que se puedan cambiar aquellos valores que no deben ser modificados.

### DIFF
--- a/FiveRockingFingers/FRF.Core/Models/Artifact.cs
+++ b/FiveRockingFingers/FRF.Core/Models/Artifact.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Xml.Linq;
 
 namespace FRF.Core.Models
@@ -14,6 +15,7 @@ namespace FRF.Core.Models
         public Project Project { get; set; }
         public int ArtifactTypeId { get; set; }
         public ArtifactType ArtifactType { get; set; }
+        public Dictionary<string, string> RelationalFields { get; set; }
 
         public virtual decimal GetPrice()
         {
--- a/FiveRockingFingers/FRF.Web.Dtos/AutoMapperProfile.cs
+++ b/FiveRockingFingers/FRF.Web.Dtos/AutoMapperProfile.cs
@@ -29,6 +29,7 @@ namespace FRF.Web.Dtos
             CreateMap<Categories.CategoryUpsertDTO, Category>();
             CreateMap<Artifact, ArtifactDTO>()
                 .ForMember(dest => dest.Settings, opt => opt.MapFrom<XmlResolver>())
+                .ForMember(dest => dest.RelationalFields, opt => opt.MapFrom<DictionaryResolver>())
                 .ForMember(dest => dest.Price, opt => opt.MapFrom(src => src.GetPrice()));
             CreateMap<ArtifactUpsertDTO, Artifact>();
             CreateMap<ArtifactType, ArtifactTypeDTO>()
--- a/FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsEc2.cs
+++ b/FiveRockingFingers/FRF.Core/Models/AwsArtifacts/AwsEc2.cs
@@ -41,6 +41,14 @@ namespace FRF.Core.Models.AwsArtifacts
         public AwsEc2(XElement settings)
         {
             Settings = settings;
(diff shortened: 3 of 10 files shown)


gperelstein (approved):

gonzalorom (approved):
That looks good for me.

FiveRockingFingers/FRF.Core/Models/Artifact.cs
```diff
@@ -14,6 +15,7 @@ public class Artifact : IArtifactPricing
         public Project Project { get; set; }
         public int ArtifactTypeId { get; set; }
         public ArtifactType ArtifactType { get; set; }
+        public Dictionary<string, string> RelationalFields { get; set; }
```
matiasbeckerle, line 18: Have you considered naming `SettingTypes` instead?

FiveRockingFingers/FRF.Web.Dtos/AutoMapperProfile.cs
```diff
@@ -29,6 +29,7 @@ public AutoMapperProfile()
             CreateMap<Categories.CategoryUpsertDTO, Category>();
             CreateMap<Artifact, ArtifactDTO>()
                 .ForMember(dest => dest.Settings, opt => opt.MapFrom<XmlResolver>())
+                .ForMember(dest => dest.RelationalFields, opt => opt.MapFrom<DictionaryResolver>())
```
matiasbeckerle, line 32: Aren't both properties a dictionary of the same type?

matiasbeckerle (approved):

FiveRockingFingers/FRF.Core/Models/Artifact.cs
martinmena: `RelationalFields` is a dictionary that contains the list of settings that can be a part of a relation, and their type. Don't know if `SettingTypes` is the best name, since there are lots of settings that can't be a part of a relation, and that name makes me think it would have all the settings and types.
Maybe `RelationalFields` isn't the best name, but I'm not quite sure in this one

FiveRockingFingers/FRF.Web.Dtos/AutoMapperProfile.cs
martinmena: Yes, but I had some issues with the content of the dictionary. For some reason, after mapping the two objects, the new dictionary was empty. Reading a few forums and questions on stack overflow, I found out that there are some problems while mapping from one dictionary to another. As far as I can tell, there is no problem working with a map from a dictionary to an object (or at least that's what the docs say), but when you map from a dictionary to another dictionary, you have to build a custom resolver or use the constructor provided by automapper, and in my opinion this was the easiest workaround.

PS: I think it has something to do with this: "_If you try create a mapper for the KeyValuePair combination you will quickly work out that you can't directly as the Key property doesn't have a setter._". (That's from one of the accepted answers in stack overflow)

FiveRockingFingers/FRF.Core/Models/Artifact.cs
matiasbeckerle: Ah gotcha, thanks for checking!

FiveRockingFingers/FRF.Web.Dtos/AutoMapperProfile.cs
matiasbeckerle: Hmm interesting. Never saw this issue before. I guess it's _yet another point against_ automapper.

PR #127, opened 2021-03-12
[FIVE-342] (Frontend) Definir settings relacionables de cada artefacto

martinmena:
## Tarea a realizar
Cada tipo de artefacto debe tener un listado de settings que pueden relacionarse, así como también que tipo de valor tienen o esperan cada una de ellas.

## Cambios realizados
Como continuación del PR #126, se modifican los formularios para creación y modificación de relaciones para mostrar solo las settings relacionables de cada artefacto, así como también el tipo de valor que representan.
_____________________________________________________________________________
#### El tipo de la setting se puede ver a la derecha del valor, entre paréntesis

image
___________________________________________________________________________
#### El listado con las settings relacionables del artefacto AmazonS3
image

### DIFF
--- a/FiveRockingFingers/FRF.Web/ClientApp/src/components/NewArtifactRelationComponents/EditArtifactRelation.tsx
+++ b/FiveRockingFingers/FRF.Web/ClientApp/src/components/NewArtifactRelationComponents/EditArtifactRelation.tsx
@@ -10,6 +10,7 @@ import ArrowBackIcon from '@material-ui/icons/ArrowBack';
 import ArrowForwardIcon from '@material-ui/icons/ArrowForward';
 import ArtifactService from '../../services/ArtifactService';
 import { Select } from '@material-ui/core';
+import { PROVIDERS } from '../../Constants';
 
 const useStyles = makeStyles((theme: Theme) =>
     createStyles({
@@ -46,8 +47,8 @@ const EditArtifactRelation = (props: { open: boolean, closeEditArtifactsRelation
 
     const classes = useStyles();
     const {handleSubmit, errors, control } = useForm();
-    const [artifact1, setArtifact1] = React.useState<Artifact | null>(props.artifactRelations.artifact1);
-    const [artifact2, setArtifact2] = React.useState<Artifact | null>(props.artifactRelations.artifact2);
+    const [artifact1, setArtifact1] = React.useState<Artifact | null>(null);
+    const [artifact2, setArtifact2] = React.useState<Artifact | null>(null);
     const [artifact1Settings, setArtifact1Settings] = React.useState<{ [key: string]: string }>({});
     const [artifact2Settings, setArtifact2Settings] = React.useState<{ [key: string]: string }>({});
     const [setting1, setSetting1] = React.useState<KeyValueStringPair | null>(null);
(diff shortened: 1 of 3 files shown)


gperelstein (approved):

FiveRockingFingers/FRF.Web/ClientApp/src/components/NewArtifactsRelation.tsx
```diff
@@ -245,6 +260,16 @@ const NewArtifactsRelation = (props: { showNewArtifactsRelation: boolean, closeN
...
+        if (!value) return "";
+
+        const result = value.replace(/([A-Z])/g, ' $1');
+        const final = result.charAt(0).toUpperCase() + result.slice(1);
+        return final;
+    }
+
     return (
         <Dialog open={props.showNewArtifactsRelation}>
             <DialogTitle id="alert-dialog-title">Formulario de para crear relación entre artefactos</DialogTitle>
```
gonzalorom, line 275: Can we change this : ***"Formulario ~~de~~ para crear...***  ?

gonzalorom (approved):
Although for me, it would be better to encapsulate some functions that are repeated in the two components, the rest looks excellent, and we can move forward.

FiveRockingFingers/FRF.Web/ClientApp/src/components/NewArtifactRelationComponents/EditArtifactRelation.tsx
```diff
@@ -183,6 +213,16 @@ const EditArtifactRelation = (props: { open: boolean, closeEditArtifactsRelation
         setRelationTypeId(event.target.value as number);
     }
 
+    const toRegularSentence = (value: string) => {
```
matiasbeckerle, line 216: Looks like you can extract this code and re-use in the other file below too.

FiveRockingFingers/FRF.Web/ClientApp/src/components/NewArtifactRelationComponents/EditArtifactRelation.tsx
```diff
@@ -183,6 +213,16 @@ const EditArtifactRelation = (props: { open: boolean, closeEditArtifactsRelation
...
+        // Value will be null only when you select a Custom artifact setting, since their type selection hasn't been implemented
+        // REMOVE ONCE IT'S DONE
+        if (!value) return "";
+
+        const result = value.replace(/([A-Z])/g, ' $1');
+        const final = result.charAt(0).toUpperCase() + result.slice(1);
```
matiasbeckerle, line 222: On one hand remember you can use string literals/interpolation. On the other hand be careful with `result.charAt(0)` or `result.charAt(0)` without any previous validation. Are you sure that `result` will always have what you expect in this other line?

matiasbeckerle (approved):

FiveRockingFingers/FRF.Web/ClientApp/src/components/NewArtifactRelationComponents/EditArtifactRelation.tsx
martinmena: We talked about this with the guys, and we think it might be best to create some "helper" classes to contain the repeated code, instead of creating the method in the parent component and drilling it down through props.
In this case, if we didn't do that, the method would end up in the component ArtifactsTable, but never be used in it, just passed down to this other two components.

This is all gonna be another ticket, since this "helpers" don't exist currently in the system (except for the HelperAddUser with some of the functionality involving users) and this means that multiple components have this problem with repeated code/code that doesn't belong to them but their children need, and we need to refactor them too.

PR #128, opened 2021-03-18
[FIVE-371]Modificar almacenamiento de sesión activa en frontend

gonzalorom:
### Descripción:
Actualmente al momento de actualizar la página o cerrar el navegador, no se persiste la sesión activa del usuario conectado. Esto se debe a que se utiliza un almacenamiento en caché a través del hook useContext. Este hook facilita enviar datos a todos los hijos del árbol de componentes, sin utilizar props. Al ser React client-side, esta información (como cualquier otro state) se elimina al actualizar el navegador.

### Modificaciones:
- Se modificó para que el perfil público del usuario conectado, se combierta a ID Token (desde el lado del servidor) y se almacene en **localstorage** o **sessionstorage** dependiendo de si al hacer login selecciono "recuérdame".
- Se creó componente de orden superior **UserContext** para que funcione junto con **useContext** y persista la sesión del usuario conectado.
- Se creó ***custom hook useUser*** para encapsular funciones y handlers relacionados al usuario, con él se puede tener acceso a logout, obtener usuario actual, almacenar datos del usuario, verificar si hay usuario está conectado y otros más.
- Se modificaron componentes `Login`, `Signup` y `NewProjectDialog` para consumir ***custom hook useUser***.


FiveRockingFingers/FRF.Web/ClientApp/src/commons/useUser.ts
```diff
@@ -0,0 +1,60 @@
...
+			}
+		} else return null;
+	};
+
+	const getUser = (): UserProfile | null => {
+		const userJson = findUser();
+		return parseUser(userJson);
+	};
+
+	const storageUser = (userJSON: string, remmember: boolean) => {
```
matiasbeckerle, line 36: Minor, I think you intended to use _"store"_ user

FiveRockingFingers/FRF.Web/ClientApp/src/commons/useUser.ts
```diff
@@ -0,0 +1,60 @@
...
+		remmember ? localStorage.setItem("user", userJSON) : sessionStorage.setItem("user", userJSON);
```
matiasbeckerle, line 37: You're not using tokens? What sort of auth mechanism you use?

matiasbeckerle (approved):

FiveRockingFingers/FRF.Web/ClientApp/src/components/auth/Signup.tsx
```diff
@@ -69,17 +68,19 @@ const Signup: React.FC<userSignUp> = ({ }) => {
...
-                    setCurrentUser(response.data);
+                    storageUser(JSON.stringify(response.data), false);
                     history.push("/Home");
                 }
             })
             .catch(function (error) {
-                if (error.response.status === 400) {
+                if (error.response.status === 401) {
+                    cleanUserStorage();
                     manageOpenSnackbar({ message: "Error al registrarse! Verifique que los datos sean correctos.", severity: "error" });
```
matiasbeckerle, line 78: Careful, 401 doesn't fit here. I think 400 was ok.

gonzalorom: I thought the same, the API response as 401, do you recommend changing the API response?

FiveRockingFingers/FRF.Web/ClientApp/src/commons/useUser.ts
gonzalorom: On the client side we don't have much auth validation. We cover the API endpoint instead with JWT made from Cognito AWS, stored on the cookies as httponly.

FiveRockingFingers/FRF.Web/ClientApp/src/commons/useUser.ts
gonzalorom: Yes you're right. Done.

FiveRockingFingers/FRF.Web/ClientApp/src/components/auth/Signup.tsx
matiasbeckerle: What I recommend is to use 401 only when it's appropriate:

_"...specifically for use when authentication is required and has failed or has not yet been provided"_

Here instead you're working with a registration endpoint. In theory, registration endpoints should be public/available since otherwise, no one will be able to create an account on the first place. That leads me to think you're not using 401 correctly in this case, probably what you intended is to use 400.

FiveRockingFingers/FRF.Web/ClientApp/src/commons/useUser.ts
matiasbeckerle: Well... I'm not an expert and I should probably read more about the subject but clients should be as lean as possible. Store the less amount of data. Only keep at hand the token to send to APIs. From that I infer that maybe you should only keep around the token and load the user data from it (or make a request to the backend) when you load your app, once, at the beginning of a page load.

I prefer not to save users data in localstorage or anywhere if possible. Not exactly sure what are the good practices today on the frontend side lately, I need to read.

FiveRockingFingers/FRF.Web/ClientApp/src/components/auth/Signup.tsx
gonzalorom: I agree, it would be different if the user was trying to sign in.  I'll make the corrections. Thanks.

FiveRockingFingers/FRF.Web/ClientApp/src/commons/useUser.ts
gonzalorom: I am not an expert either (I am very, very far from it). What I can say about what I have been researching and reading, is that it is a common practice to store certain pieces of information in the browser (the same happens in mobile apps).

Let's take ***GitHub*** as an example, if we count the 🍪 that it stores, we have a total of ***10 (ten)*** and if we count the information stored in LocalStorage, we find ***3 (three)*** of which several are arrays of complex objects.

I will investigate more on the subject. For now, I am going to encrypt the information we store so that it is not so human friendly.

matiasbeckerle: Yep I'm not saying to not use localstorage at all. Just to try as much as possible to avoid storing _user_ info. Sometimes by storing the short-living tokens is enough and I was wondering if that is a possibility here. As an example, some tokens contains some user data too, like a display name and maybe an email. If that's the case, you only keep around the token, and one the webapp loads, you read from token the core user data. Just an idea, there are probably better. My intention with this is that could can have it in mind as a team and check for other options or confirm this was the best way (not today, with time).

gonzalorom: We will go with that approach, just storing the short-living token and implement a JWT decoder to get the user info from it. The info that we are already sending is the public user profile (name, email, avatar). Thank you so much for all the feedback.

matiasbeckerle: No, thank you. If by any chance you keep reading about this subject and found interesting articles please share!

gonzalorom: Done! You can find the implementation here: #130

martinmena (approved):

gperelstein (approved):

PR #143, opened 2021-03-31
[FIVE-343] (FrontEnd) Crear control de tipo en el valor de las settings y relaciones

martinmena:
## Tarea a realizar
_Se debe modificar la creación de relaciones para solo permitir crear relaciones entre settings que esperen y aporten el mismo tipo de valor (un entero a una setting que espera un entero por ejemplo)._

## Cambios realizados
Tanto en modificación como en creación de settings, al seleccionar las dos settings se verifica que ambas sean del mismo tipo. En caso de que no, los campos correspondientes del formulario pasan al color rojo, y no se permite continuar con la acción (creación o modificación) hasta solucionarlo.

Además, se crea HelperArtifactRelation con la funcionalidad compartida entre los componentes de creación y de modificación de relaciones.

#### Dos settings del mismo tipo (se puede ver su tipo entre paréntesis debajo del nombre de las settings)
image

#### Dos settings de diferente tipo
image

### DIFF
--- a/FiveRockingFingers/FRF.Web/ClientApp/src/components/NewArtifactRelationComponents/EditArtifactRelation.tsx
+++ b/FiveRockingFingers/FRF.Web/ClientApp/src/components/NewArtifactRelationComponents/EditArtifactRelation.tsx
@@ -10,7 +10,7 @@ import ArrowBackIcon from '@material-ui/icons/ArrowBack';
 import ArrowForwardIcon from '@material-ui/icons/ArrowForward';
 import ArtifactService from '../../services/ArtifactService';
 import { Select } from '@material-ui/core';
-import { PROVIDERS } from '../../Constants';
+import { updateArtifactsSettings, toRegularSentence, handleArtifactChange, handleSettingChange, differentSettingTypes } from './HelperArtifactRelation';
 
 const useStyles = makeStyles((theme: Theme) =>
     createStyles({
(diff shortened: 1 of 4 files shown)


FiveRockingFingers/FRF.Web/ClientApp/src/components/NewArtifactRelationComponents/HelperArtifactRelation.ts
```diff
@@ -0,0 +1,111 @@
...
+import KeyValueStringPair from "../../interfaces/KeyValueStringPair";
+
+export function updateArtifactsSettings(
+    artifact1: Artifact | null,
+    artifact2: Artifact | null,
+    setArtifact1Settings: Function,
+    setArtifact2Settings: Function
+) {
+    if (artifact1 !== null && artifact1 !== undefined) {
+        if (artifact1.artifactType?.name != PROVIDERS[1]) {
```
matiasbeckerle, line 12: This code looks quite tied to AWS provider, isn't it? Is that what `PROVIDERS[1]` means? If that's the case maybe you will want to make the code as agnostic possible so if the time to add a different provider comes in, the amount of changes will be way less. Since the project it's still in kinda of a demo/prototype/vertical-slice by using only AWS I guess that's fine anyways to move forward as you folks did in other places. If this is not about AWS specific code then sorry for the preach 😅

matiasbeckerle (approved):

FiveRockingFingers/FRF.Web/ClientApp/src/components/NewArtifactRelationComponents/HelperArtifactRelation.ts
martinmena: I would say it's the other way around. `PROVIDERS[1] = 'Custom'`, so every provider would have the same behavior except the Custom one.

FiveRockingFingers/FRF.Web/ClientApp/src/components/NewArtifactRelationComponents/EditArtifactRelation.tsx
```diff
@@ -10,7 +10,7 @@ import ArrowBackIcon from '@material-ui/icons/ArrowBack';
 import ArrowForwardIcon from '@material-ui/icons/ArrowForward';
 import ArtifactService from '../../services/ArtifactService';
 import { Select } from '@material-ui/core';
-import { PROVIDERS } from '../../Constants';
+import { updateArtifactsSettings, toRegularSentence, handleArtifactChange, handleSettingChange, differentSettingTypes } from './HelperArtifactRelation';
```
gonzalorom, line 13: Well done mate! Excellent job with the `HelperArtifactRelation`.

gonzalorom (approved):

FiveRockingFingers/FRF.Web/ClientApp/src/components/NewArtifactRelationComponents/HelperArtifactRelation.ts
matiasbeckerle: Nice! Since the custom one is special, maybe you could have another way to fetch it instead of PROVIDERS[1]? Maybe as CUSTOM_PROVIDER?

martinmena: What about something like this?

Instead of (as it is right now)
`export const PROVIDERS = ['AWS', 'Custom'];`

Doing something like this, and every time we add a new provider, we create a new const, and add it to the array. That way, everything still works the same, but we have a more specific way to access each one of them 
```
export const CUSTOM_PROVIDER = 'Custom';
export const AWS_PROVIDER = 'AWS';

export const PROVIDERS = [AWS_PROVIDER, CUSTOM_PROVIDER];
```

matiasbeckerle: Sounds good to me, thanks!

gperelstein (approved):

PR #149, opened 2021-04-08
[FIVE-395] Modificar respuestas de controllers integrando ServiceResponse

gonzalorom:
### Algunos ejemplos antes de la modificación:
viejoErrorUpdate
viejoError
### Ahora:
nuevoError2
nuevoError

- Se modificó ErrorCodes RelationNotValid para evitar generalizar error cuando las relaciones son de diferente tipo, están repetidas  o tienen diferente artefacto base.
- Se agregó data annotation para validar que la lista de usuarios al actualizar un proyecto contenga al menos un usuario.
- Se modificaron respuestas de ArtifactTypesController, BudgetController, CategoriesController, SignInController, SignUpController y UserController para que implementen ErrorCode y error message de ServiceResponse.
- Se agregaron y acomodaron Unit Test.

### DIFF
--- a/FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
+++ b/FiveRockingFingers/FRF.Core/Response/ErrorCodes.cs
@@ -12,9 +12,11 @@
 
         // Relation errors
         public const int RelationAlreadyExisted = 10;
-        public const int RelationNotValid = 11;
-        public const int RelationNotExists = 12;
-        public const int RelationCycleDetected = 13;
+        public const int RelationNotValidDifferentBaseArtifact = 11;
+        public const int RelationNotValidRepeated = 12;
+        public const int RelationNotValidDifferentType = 13;
+        public const int RelationNotExists = 14;
+        public const int RelationCycleDetected = 15;
 
         // User errors
         public const int InvalidCredentials = 20;
--- a/FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
+++ b/FiveRockingFingers/FRF.Core/Services/ArtifactsService.cs
@@ -406,17 +406,17 @@ namespace FRF.Core.Services
 
             var hasAnyRelationWithoutBaseArtifact = HasAnyRelationWithoutBaseArtifact(artifactId, artifactRelations);
             if (hasAnyRelationWithoutBaseArtifact)
-                return new ServiceResponse<IList<ArtifactsRelation>>(new Error(ErrorCodes.RelationNotValid,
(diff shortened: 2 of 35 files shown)


martinmena:
I'm not sure you should be returning the error number in the controller response, since it's a specific codification for our error system. It's something internal to help us, not the user.

gonzalorom:
> 
> 
> I'm not sure you should be returning the error number in the controller response, since it's a specific codification for our error system. It's something internal to help us, not the user.
> 
> (I could be completely wrong with this comment)

I see this situation somewhat common on several websites, where one receives an error number (other than an HTTP Status code) along with the message. Facebook or Jira API itself perform this same action. 

Perhaps if this API go public, along with the documentation the IDs of those specific errors can be exposed so that the developer implementing the API can make use of them. Even in some cases a URL with more information about it is usually included. For all this, it does not seem a bad idea to only expose them when one makes a call to the endpoint. 

PS: An article that I find very interesting, that talks about this topic is the following: https://blog.restcase.com/rest-api-error-codes-101/

matiasbeckerle:
I hope this can clarify a little. In general I agree with @gonzalorom's latest comment but it's worth nothing a couple of things.

- APIs should not be (as @martinmena pointed out) **end user** friendly. Notice the _end_. Anything we expose as API builders shall not be used for applications as direct output to end users. If you expose an error on the API and the consumer of the API doesn't do anything and just print the result... well... bad experience. That's not very good. Instead, the API consumer needs to build around what we have returned. In this case, interpreting the error number and display something nice or helpful to the end user.
- With these changes we are loosing something important: the error object. API consumers (in your case, the React app) will no longer be able to interpret the error code because everything is just a string. A string that starts with the number and a brief description. I recommend restoring back the error number on one hand and the description on the other.

gonzalorom:
> 
> 
> I hope this can clarify a little. In general I agree with @gonzalorom's latest comment but it's worth nothing a couple of things.
> 
>     * APIs should not be (as @martinmena pointed out) **end user** friendly. Notice the _end_. Anything we expose as API builders shall not be used for applications as direct output to end users. If you expose an error on the API and the consumer of the API doesn't do anything and just print the result... well... bad experience. That's not very good. Instead, the API consumer needs to build around what we have returned. In this case, interpreting the error number and display something nice or helpful to the end user.
> 
>     * With these changes we are loosing something important: the error object. API consumers (in your case, the React app) will no longer be able to interpret the error code because everything is just a string. A string that starts with the number and a brief description. I recommend restoring back the error number on one hand and the description on the other.

Thanks for the feedback, well I'll change the response to an object with both error number and description and make the corresponding changes in the front.

gperelstein (approved):

martinmena (approved):

